PR #31, opened 2019-09-11
Solution

PastushenkoVik:
https://pastushenkovik.github.io/react_list-of-goods/

### DIFF
--- a/src/components/Goods/Goods.js
+++ b/src/components/Goods/Goods.js
@@ -59,6 +59,12 @@ export default class Goods extends Component {
 
   render() {
     const { goods } = this.props;
+
+    const SELECT_LENGTH = 10;
+    const selectList = [...Array(SELECT_LENGTH + 1).keys()]
+      .slice(1)
+      .map(item => ({ name: item, val: item }));
+
     let goodsList = [...goods];
 
     if (this.state.select.isChange) {
@@ -87,7 +93,7 @@ export default class Goods extends Component {
         </div>
         <div>
           <Select
-            rows={10}
+            list={selectList}
             selected={this.state.select.value}
             onChange={this.onChangeSelect}
           />
--- a/src/components/Goods/Goods.js
+++ b/src/components/Goods/Goods.js
@@ -25,13 +25,11 @@ export default class Goods extends Component {
   }
 
   onClickReverse() {
-    // eslint-disable-next-line react/no-access-state-in-setstate
-    this.setState({ isReverse: !this.state.isReverse });
+    this.setState(prevState => ({ isReverse: !prevState.isReverse }));
   }
 
   onClickAlphabetical() {
-    // eslint-disable-next-line react/no-access-state-in-setstate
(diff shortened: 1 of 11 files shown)


src/components/Goods/Goods.js
```diff
@@ -0,0 +1,77 @@
...
+export default class Goods extends Component {
+  state = {
+    goods: [...this.props.goods],
+    selectValue: 1,
+    selectList: [...Array(SELECT_LENGTH + 1).keys()]
+      .slice(1)
+      .map(item => ({ name: item, value: item })),
+  };
+
+  reverseClick = () => {
```
irenhh: change the name according to Yura's options, current name isn't good

src/components/Goods/Goods.js
```diff
@@ -0,0 +1,77 @@
...
+    this.setState(prevState => ({ goods: prevState.goods.reverse() }));
+  };
+
+  alphabeticalClick = () => {
+    this.setState(prevState => ({ goods: prevState.goods.sort() }));
+  };
+
+  resetClick = () => {
```
irenhh: the same with name,
it seems that you're trying to reset click, but actually you need to reset state of items list

src/constants/proptypes.js
```diff
@@ -0,0 +1,25 @@
...
+  list: PropTypes.arrayOf(PropTypes.objectOf(SelectShape)),
+  selected: PropTypes.number.isRequired,
+  onChange: PropTypes.func.isRequired,
+};
+
+export const GoodTypes = {
+  good: PropTypes.string.isRequired,
+};
+
+export const GoodsTypes = {
```
irenhh: this name and the prev one are very similar, it can happen that somebody doesn't notice this and thinks they're the same

src/components/Goods/Goods.js
```diff
@@ -0,0 +1,77 @@
...
+import Good from '../Good/Good';
+import Button from '../Button/Button';
+import Select from '../Select/Select';
+
+const SELECT_LENGTH = 10;
+
+export default class Goods extends Component {
+  state = {
+    goods: [...this.props.goods],
+    selectValue: 1,
```
irenhh: `selectedValue`, 
current name is for functions since it describes action

src/components/Goods/Goods.js
```diff
@@ -0,0 +1,77 @@
...
+    selectList: [...Array(SELECT_LENGTH + 1).keys()]
```
irenhh: the same here

src/components/Goods/Goods.js
```diff
@@ -0,0 +1,77 @@
+import React, { Component } from 'react';
+import './Goods.scss';
+import { GoodsTypes } from '../../constants/proptypes';
+import Good from '../Good/Good';
+import Button from '../Button/Button';
+import Select from '../Select/Select';
+
+const SELECT_LENGTH = 10;
```
irenhh: you missed Yura's comment about this

PR #115, opened 2020-02-17
solution

Annasmile3:
[DEMO LINK](https://Annasmile3.github.io/react_list-of-goods/)


src/App.css
```diff
@@ -0,0 +1,51 @@
+.App {
```
danheim:
```suggestion
.app {
```

PR #137, opened 2020-02-22
List_of_goods

delenjer:
[DEMO LINK](https://delenjer.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { GoodsList } from './components/GoodsList';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -14,11 +15,126 @@ const goodsFromServer = [
...
+        visibleGoods: [...prevState.visibleGoods].sort(
+          (a, b) => a.length - b.length,
+        ),
+      }
+    ));
+  }
+
+  onSelectChange = (event) => {
+    this.setState({ selectedValue: event.target.value });
+    this.setState(prevState => (
```
irenhh: this is a bad idea to call setState function several times, it's async function

irenhh:
```
this.setState((prevState) => ({
  selectedValue: event.target.value,
  visibleGoods: .....,
}))
```

PR #182, opened 2020-05-05
Nina

nchuhrina:
[DEMO LINK](https://nchuhrina.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import Action from './components/Action';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -14,11 +15,26 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+// eslint-disable-next-line react/prefer-stateless-function
```
irenhh, line 18: do you really need this?

src/App.js
```diff
@@ -14,11 +15,26 @@ const goodsFromServer = [
...
+class App extends React.Component {
+  state ={
```
irenhh, line 20:
```suggestion
  state = {
```

src/components/Action.js
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class Action extends React.Component {
+  state = {
+    goodsLength: 1,
+    goods: [...this.props.goods],
```
irenhh, line 7: why don't you import goods in this file? no need to pass them as props

src/components/Action.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  reverse = () => {
+    this.setState(prevState => ({
+      goods: [...prevState.goods]
+        .reverse(),
+    }));
+  }
+
+  sortAlphabetically = () => {
```
irenhh, line 17: use `localeCompare()` for alphabetical sorting

src/components/Action.js
```diff
@@ -0,0 +1,89 @@
...
+          </button>
+          <button type="button" onClick={this.sortByLength}>
+            Sort by length
+          </button>
+
+          <select
+            onInput={this.selectChange}
+            defaultValue={this.state.goodsLength}
+          >
+            {[...Array(10).keys()].map((item, index) => (
```
irenhh, line 75: move this array outside of render, you don't need to recreate it on each render, it is always the same

irenhh:
if I sort alphabetically and then reduce word length, sorting should be still enabled

PR #225, opened 2020-09-03
Solution

uni-8:
- [DEMO LINK](https://uni-8.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,7 @@
 import React from 'react';
+import { GoodsList } from './components/GoogsList';
+import { Button } from './components/Button';
+import { Select } from './components/Select';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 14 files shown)


src/App.js
```diff
@@ -14,11 +17,94 @@ const goodsFromServer = [
...
+  reverseList = () => (
+    this.setState(state => ({
+      goods: [...state.goods].reverse(),
+    }))
+  );
+
+  sortAlphabetically = () => (
+    this.setState(state => ({
+      goods: [...state.goods].sort(),
+    }))
```
tykhan, line 43: for alphabetically sorting is a little better to use localeCompare

tykhan (approved):
nice👍

src/App.js
irenhh: why didn't you fix this?

src/App.js
```diff
@@ -14,11 +17,94 @@ const goodsFromServer = [
...
+    const { wasStarted, goods, wordLength } = this.state;
+
+    return (
+      <div className="App">
+        <h1 className="App__header">Goods</h1>
+
+        <div>
+          <Button
+            wasStarted={!wasStarted}
+            call={this.reset}
```
irenhh, line 64: what does call mean here?

src/App.js
```diff
@@ -14,11 +17,94 @@ const goodsFromServer = [
...
+            call={this.selectByLength}
+            targetLength={+wordLength}
+            initList={goodsFromServer}
+          />
+        </div>
+
+        <GoodsList
+          wasStarted={!wasStarted}
+          goods={goods}
+          targetLength={+wordLength}
```
irenhh, line 103: isn't it a number already?

src/App.js
uni-8: cuz i agree that localeCompare is a best practice, but in this task, it was not necessary. i fixed it now

src/App.js
uni-8: yes, it was

src/App.js
uni-8: function call. i changed this to action. is it better now?

Denchepornyuk (approved):

PR #322, opened 2021-01-19
List of goods solution

VinnieJ-2k20:
- [DEMO LINK](https://VinnieJ-2k20.github.io/react_list-of-goods/)

Advanced part will be done later.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { GoodsList } from './GoodsList';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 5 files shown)


olesiatk (approved):
Great!

sasha39612 (approved):

src/App.js
```diff
@@ -14,11 +15,126 @@ const goodsFromServer = [
...
+          ))}
+        </select>
+
+        {isContentVisible
+          && (
+            <GoodsList
+              goods={goodsFromServer}
+              reversed={isReversed}
+              sortedBy={sortedBy}
+              minLength={lengthFilter}
```
varseniuk, line 134: Let's pass only prepared goods to the component GoodsList. It should only render the list, move all your operations (sorting, reversing etc) to the parent element and pass here only ready goods

src/GoodsList.jsx
```diff
@@ -0,0 +1,41 @@
...
+    case 'length':
+      processedList.sort((a, b) => a.length - b.length);
+      break;
+
+    default:
+  }
+
+  if (reversed) {
+    processedList.reverse();
+  }
```
varseniuk, line 23: Let's move it to the parent element

varseniuk (changes requested):
Everything is very good, but let's make small refactoring

varseniuk (approved):

PR #325, opened 2021-01-19
Develop

maryna-zinchenko:
[DEMO LINK](https://maryna-zinchenko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,6 @@
 import React from 'react';
+import classNames from 'classnames';
+import { GoodsList } from './components/GoodsList';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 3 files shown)


Pasha5555 (approved):
Good job!

src/App.js
```diff
@@ -14,11 +16,114 @@ const goodsFromServer = [
...
-const App = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.PureComponent {
+  state = {
+    hideButton: false,
+    showList: true,
```
varseniuk, line 22: Let's refactor this a little bit and don't use 'hidden' attribute. Let's use conditions rendering instead.
I mean that initially 'showList' should be 'false' and you should render your button according to that variable:
`{!showList && <button>Start</button>}`

by clicking this button you change your state's 'showList' to 'true' what will automatically hide the button and according to this variable render everything else:

```
{showList && (
  <GoodsList />
  ...
  <div>...</div>
  ...
  <button> ...</button>
  **and everything else here**
)}
```

src/App.js
```diff
@@ -14,11 +16,114 @@ const goodsFromServer = [
...
+    return (
+      <div className="App box has-text-centered">
+        <h1 className="title is-1">Goods</h1>
+        <GoodsList
+          showList={showList}
+          goods={goodsFromServer}
+          isReversed={isReversed}
+          sortBy={sortBy}
+          value={value}
+        />
```
varseniuk, line 71: You should do all the operations (sorting, reversing etc.) on this level (in the parent element) and pass only preparedGoods to 'GoodsList' component, which should only render the list with already prepared (sorted or reversed or whatever) goods

src/App.js
```diff
@@ -14,11 +16,114 @@ const goodsFromServer = [
...
+            <option value="2">2</option>
+            <option value="3">3</option>
+            <option value="4">4</option>
+            <option value="5">5</option>
+            <option value="6">6</option>
+            <option value="7">7</option>
+            <option value="8">8</option>
+            <option value="9">9</option>
+            <option value="10">10</option>
+          </select>
```
varseniuk, line 122: It would be better to render this 'select' using an array of values in state and mapping it (like you do with lists).

src/components/GoodsList.jsx
```diff
@@ -0,0 +1,47 @@
...
+          return g1.length - g2.length;
+
+        default:
+          return 0;
+      }
+    });
+
+    if (isReversed) {
+      goodsList.reverse();
+    }
```
varseniuk, line 29: Let's move it to the parent element

varseniuk (changes requested):
Let's refactor this a bit

varseniuk (approved):

PR #327, opened 2021-01-19
solution

zhenya-mezhueva:
[DEMO LINK](https://zhenya-mezhueva.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import GoodsList from './GoodsList';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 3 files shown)


maryna-zinchenko (approved):
good job!

src/App.js
```diff
@@ -14,11 +15,90 @@ const goodsFromServer = [
...
+            }
+          </button>
+          <div className="goods-block">
+            {isListVisible && (
+              <GoodsList
+                goods={goods}
+                reverse={this.reverse}
+                sortByAlphabet={this.sortByAlphabet}
+                sortByLength={this.sortByLength}
+                reset={this.reset}
```
vsm2k17, line 94: if you already sort an array why do you need to pass all those props, I think goods should enough.

src/GoodsList.js
```diff
@@ -0,0 +1,53 @@
...
+        type="button"
+        className="btn"
+        onClick={sortByAlphabet}
+      >
+        Sort A-Z
+      </button>
+      <button
+        type="button"
+        className="btn"
+        onClick={sortByLength}
```
vsm2k17, line 30: I think buttons should be outside of the goodsList component

vsm2k17 (changes requested):

src/GoodsList.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './App.css';
+
+const GoodsList = ({ goods }) => (
+  <>
```
vsm2k17, line 6: no need to use react.fragment you already wrap everything in ul tag

vsm2k17 (approved):
good job

PR #329, opened 2021-01-19
Solution

AlexysUa:
[DEMO LINK](https://alexysua.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,8 @@
 import React from 'react';
 import './App.css';
+import 'bootstrap/dist/css/bootstrap.min.css';
+import { Button } from 'react-bootstrap';
+import ProductList from './components/ProductList';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -13,12 +16,138 @@ const goodsFromServer = [
...
+    this.setState(prevState => ({
+      ...prevState,
+      products: [...prevState.products].sort(),
+    }));
+  }
+
+  reset = () => {
+    this.setState(prevState => ({
+      ...prevState,
+      products: initialState.products,
```
MarchD, line 54: I think we should try use state instead of variables which are initialized outside

MarchD (approved):
good job )

src/App.js
AlexysUa: Fixed!

src/App.js
```diff
@@ -14,11 +17,137 @@ const goodsFromServer = [
...
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    listIsVisible: true,
+    buttonIsVisible: false,
+    products: goodsFromServer,
+    initialState: goodsFromServer,
```
vsm2k17, line 25: no need to save it here, you can access them anyway

src/App.js
```diff
@@ -14,11 +17,137 @@ const goodsFromServer = [
...
+    selectedElementLength: 1,
+  }
+
+  showList = () => {
+    this.setState(prevState => ({
+      ...prevState,
+      listIsVisible: !prevState.listIsVisible,
+      buttonIsVisible: !prevState.buttonIsVisible,
```
vsm2k17, line 33: no need to have two properties for this, if the list is visible, that means the button is hidden, so one property should be enough

src/App.js
```diff
@@ -14,11 +17,137 @@ const goodsFromServer = [
...
+    this.setState(prevState => ({
+      ...prevState,
+      products: [...prevState.products].sort(),
+    }));
+  }
+
+  reset = () => {
+    this.setState(prevState => ({
+      ...prevState,
+      products: prevState.initialState,
```
vsm2k17, line 54:
```suggestion
      products: goodsFromSerever,
```

src/App.js
```diff
@@ -14,11 +17,137 @@ const goodsFromServer = [
...
+    this.setState(prevState => ({
+      products: prevState.initialState
+        .filter(product => product.length >= target.value),
+    }));
+  }
+
+  render() {
+    const { listIsVisible, products, buttonIsVisible } = this.state;
+    const productsCopy = [...products];
+    const selectArray = new Array(10).fill(0);
```
vsm2k17, line 76: create array outside of the component, no need to create It every time when your component is re-rendering

src/App.js
```diff
@@ -14,11 +17,137 @@ const goodsFromServer = [
...
+            <select
+              name="selectList"
+              id="selectList"
+              size={3}
+              onChange={this.sortBySelect}
+            >
+              {selectArray.map((select, index) => (
+                <option
+                  value={index + 1}
+                  key={Math.random()}
```
vsm2k17, line 96: you can create an array with indexes as elements instead of an array with zeros, and use these elements as keys.

src/components/ProductList/ProductList.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import './ProductList.css';
+import 'bootstrap/dist/css/bootstrap.min.css';
+
+export const ProductList = ({ products }) => (
+  <>
```
vsm2k17, line 6: redundant

src/App.js
```diff
@@ -14,11 +17,137 @@ const goodsFromServer = [
...
+          </Button>
+          <Button
+            variant="info"
+            type="button"
+            onClick={this.sortByLength}
+          >
+            Sort by length
+          </Button>
+          <div hidden={listIsVisible}>
+            <ProductList
```
vsm2k17, line 143: use conditional rendering

vsm2k17 (changes requested):
image
your select a bit broken, really hard to click on it

src/App.js
AlexysUa: FIxed!

src/components/ProductList/ProductList.jsx
vsm2k17: you already wrap everything in ul tag, it can be removed

vsm2k17 (approved):

PR #331, opened 2021-01-19
solution

Pasha5555:
[DEMO LINK](https://pasha5555.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,7 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
+
+import GoodsList from './components/GoodsList';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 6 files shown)


maryna-zinchenko (approved):

src/App.js
```diff
@@ -14,11 +16,127 @@ const goodsFromServer = [
...
+                  <option value="2">2</option>
+                  <option value="3">3</option>
+                  <option value="4">4</option>
+                  <option value="5">5</option>
+                  <option value="6">6</option>
+                  <option value="7">7</option>
+                  <option value="8">8</option>
+                  <option value="9">9</option>
+                  <option value="10">10</option>
+                </select>
```
varseniuk, line 118: Better to create an array with options ([1,2,3,4, ... 10]) and render it here using 'map' (like you're doing with lists).
Also, the 'vale' attribute should be synchronized with 'selectedValue' in your state (`<option value="this.state. selectedValue">`).
It's important here because when you resetting everything (using your 'reset' method) the values of his select stay unchangeable. After clicking 'reset' the value of the select should also be resetted

src/App.js
```diff
@@ -14,11 +16,127 @@ const goodsFromServer = [
...
+                >
+                  Reset
+                </button>
+
+                <GoodsList
+                  goods={goods}
+                  isReversed={isReversed}
+                  sortBy={sortBy}
+                  selectedValue={selectedValue}
+                />
```
varseniuk, line 133: You should transfer to this component only 'goods'. They should be sorted and reversed here, and only after that, all changed goods should be transferred to 'GoodsList'.
All operations are made on this level, and component 'GoodsList' should only render a list

varseniuk (changes requested):

varseniuk (approved):

PR #339, opened 2021-01-20
solution

HaidaiAndrii:
[DEMO LINK](https://HaidaiAndrii.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import { GoodsList } from './Components/GoodsList/GoodsList';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -14,11 +15,114 @@ const goodsFromServer = [
...
+          </button>
+        )
+        }
+        {isVisible
+          && (
+            <>
+              <GoodsList
+                goods={copyGoods}
+                sortedBy={sortedBy}
+                isReversed={isReversed}
```
vsm2k17, line 80: just pass already prepared array, no need to do such things inside GoodsList component

vsm2k17 (changes requested):

src/App.js
vsm2k17: remove props that you don't use

vsm2k17 (changes requested):

vsm2k17 (approved):
good job

PR #340, opened 2021-01-20
solution

Ruslanwp:
[DEMO LINK](https://Ruslanwp.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import './App.css';
+import { Goods } from './components/Goods';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


olesiatk (approved):
Good!

src/App.js
```diff
@@ -14,11 +14,110 @@ const goodsFromServer = [
...
-const App = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    goods: goodsFromServer,
+    startButton: false,
```
varseniuk, line 20: It's a bad name and it's hard to understand for what purposes is used.
This variable actually represents if the list is visible or not, so it would be better to call it smth like 'isListVisible'

varseniuk (approved):
Good, but pay attention to naming of variables

PR #344, opened 2021-01-20
List of Goods with with list control buttons on React

KirillLutsenko:
[DEMO LINK](https://KirillLutsenko.github.io/react_list-of-goods/)


taniazt (approved):

src/components/GoodsList/GoodsList.jsx
```diff
@@ -0,0 +1,122 @@
...
+          </div>
+        </div>
+
+        <ul className="goods-list">
+          {goods.map(good => (
+            <li className="goods-list__item" key={good}>
+              {good}
+            </li>
+          ))}
+        </ul>
```
varseniuk, line 114: Your component 'GoodsList' should only render the list with ALREADY prepared (reverse, sorted, etc) goods.
So you should write all the logic, all operations, and all methods on the upper level, then pass prepared goods to 'GoodsList' component and render them here.

src/App.jsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  render() {
+    const { isStartButtonVisible } = this.state;
+
+    return (
+      <div className="app">
+        {isStartButtonVisible
+          ? <Start showGoodsList={this.showGoodsList} />
+          : <GoodsList goods={goodsFromServer} />
```
varseniuk, line 37: Let's split the logic. Make all operations here and pass already transformed goods from server to the 'GoodsList' component

src/components/GoodsList/GoodsList.jsx
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+import './GoodsList.css';
+import PropTypes from 'prop-types';
+
+const selectItems = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
```
varseniuk, line 5: For this task, it's ok, but better not to hardcode values and use loops, or array methods for filling the arrays with some values

varseniuk (changes requested):

varseniuk (approved):
👍

PR #352, opened 2021-01-20
reactive list of goods

WhiteRhin0:
- [DEMO LINK](https://WhiteRhin0.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { GoodsList } from './components/GoodsList';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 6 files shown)


zhenya-mezhueva (approved):

src/App.js
```diff
@@ -14,11 +15,103 @@ const goodsFromServer = [
...
+  alphabetSort = () => {
+    this.setState({ sortedBy: 'alphabet' });
+  }
+
+  resetIt = () => {
+    this.setState({
+      sortedBy: '',
+      isReversed: false,
+    });
+  }
```
varseniuk, line 49: Let's make this function do what it actually should do.
If it's called 'reset' so it should return the original list (not sorting 'by nothing', as you have written).
You can just assign the original 'goodsFromServer' to your state here

src/components/GoodsList/GoodsList.jsx
```diff
@@ -0,0 +1,18 @@
...
+export const GoodsList = ({ goods }) => (
+  <>
+    <ul>
+      {goods.map(good => (
+        <li key={good}>
+          {good}
+        </li>
+      ))}
+    </ul>
+  </>
```
varseniuk, line 13:
```suggestion
    <ul>
      {goods.map(good => (
        <li key={good}>
          {good}
        </li>
      ))}
    </ul>
  ```

No need in the fragment here, Everything is already wrapped in 'ul' tag and it is the only tag here

varseniuk (changes requested):

src/App.js
WhiteRhin0: добавил строчку "goods: goodsFromServer", хотя не уверен, что это как-то влияет. А если удалить отсюда смену состояния sortedBy и isReversed, то тогда будут выполняться блоки if в render(), и reset не работает. Поэтому и подумал, что reset должен сбрасывать все состояния

src/components/GoodsList/GoodsList.jsx
WhiteRhin0: done

src/App.js
varseniuk: You correct about sortedBy and isReversed. You should reset them too. What I was telling you that semantically method 'reset' should revert everything to its initial state, but not sort thing by nothing.
When you write `goods: goodsFromServer` that's mean that you putting original goodsFromServer into your state. Which means that everything goes back to where it began. But when you trying to sort it by nothing, it's not quite understandable for smbd else for which purposes you make 'sort' inside method which is called 'reset'

varseniuk (approved):
👍

PR #353, opened 2021-01-20
react_list-of-goods

Nvalja:
[DEMO LINK](https://Nvalja.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,7 @@
 import React from 'react';
-import './App.css';
+import classNames from 'classnames';
+import './App.scss';
+import { GoodsList } from './components/GoodsList';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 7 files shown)


kerpichonka (approved):
awesome!

src/App.js
```diff
@@ -14,11 +16,110 @@ const goodsFromServer = [
...
+  isStarted = () => {
+    this.setState({ startedList: true });
+  };
+
+  resetList = () => {
+    this.setState({
+      sortBy: '',
+      reversed: false,
+    });
+  };
```
varseniuk, line 36: Let's make this function do what it actually should do.
If it's called 'reset' so it should return the original list (not sorting 'by nothing', as you have written).
You can just assign the original 'goodsFromServer' to your state here

src/App.js
```diff
@@ -14,11 +16,110 @@ const goodsFromServer = [
...
+    }
+
+    return (
+      <div className="App">
+        <h1 className="App__title">Goods</h1>
+        <button
+          type="button"
+          className={classNames(
+            'App__button',
+            { 'App__button--hidden': startedList },
```
varseniuk, line 78: Just FYI: you could use conditional rendering for this button instead of 'hidden' attribute

`{!startedList && <button> ... </button>}`

varseniuk (changes requested):
Let's make some small fixes

src/App.js
Nvalja: done)

src/App.js
Nvalja: 👍

varseniuk (approved):
Good job 👍

PR #368, opened 2021-02-27
Develop

oleksandrorel:
- [DEMO LINK](https://oleksandrorel.github.io/react_list-of-goods/)
- Created class App that has `start`, `goods`, `goodsLength` state properties `showList`, `reverse`, `alphabeticalSort`, `lengthSort`, `reset`, `getSelectedLength` `render` methods.
- `showList` method set `start` state property as `true`
- `reverse` method creates copy of goods array that stored in `state` and reverse it. 
- `alphabeticalSort` method creates copy of goods array that stored in `state` and sort it in alphabet way. 
- `lengthSort` method creates copy of goods array that stored in `state` and sort goods ordered by name length.  
- `reset` method set initial array of goods `goodsFromServer` to state array `goods` and set state property `goodsLength` as `1`.
- `getSelectedLength` method set selected value by user to `goodsLength` state property.
- Method `render has `visibleGoods` array that stored filtered goods based on the length of goods name characters, respective click handlers on buttons that runs corresponding methods. Method call component `GoodsList` and pass  `visibleGoods` array as parameter. Method has `select` form that allows user to set the number of good characters as a filter criterion and will be passed to `getSelectedLength` method. Method also returns respective layout of the App.
- Component `GoodsList` accept `visibleGoods` array and returns layout of goods list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # React list of goods
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_list-of-goods/)
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://oleksandrorel.github.io/react_list-of-goods/)
 - Follow the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline)
 
 ## Tasks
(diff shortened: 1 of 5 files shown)


vpolets (approved):
Класно зробив і дизайн подобається)

PR #431, opened 2021-05-23
Solution

Burunduk003:
- [DEMO LINK](https://Burunduk003.github.io/react_list-of-goods/)

### DIFF
--- a/src/components/GoodsList/GoodsList.js
+++ b/src/components/GoodsList/GoodsList.js
@@ -1,0 +1,158 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import './GoodsList.css';
+
+export class GoodsList extends Component {
+  state ={
+    hidden: true,
+    reverse: false,
+    sortBy: '',
+    goodsSelector: 1,
+  }
+
+  hideButton = () => {
+    this.setState({
+      hidden: false,
+    });
+  }
+
+  reverseGoods = () => {
+    this.setState(state => ({
+      reverse: !state.reverse,
+    }));
+  }
+
+  sortGoods = () => {
+    this.setState(state => ({
+      sort: !state.sort,
+    }));
+  }
+
+  resetGoods = () => {
+    this.setState({
+      reverse: false,
+      sortBy: '',
+      goodsSelector: 1,
+    });
+  }
+
+  render() {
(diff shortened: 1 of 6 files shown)


Ihor-Karpyn (approved):
Good job!

src/components/GoodsList/GoodsList.js
```diff
@@ -0,0 +1,150 @@
...
+      <>
+        <button
+          type="button"
+          className={classNames('GoodsList__toggleButton', {
+            'GoodsList__toggleButton--hide': !hidden,
+          })}
+          onClick={this.hideButton}
+        >
+          Start
+        </button>
```
mykola-yarchuk, line 72: hide this button with conditional rendering, no need to use css at all

src/components/GoodsList/GoodsList.js
```diff
@@ -0,0 +1,150 @@
...
+        <div
+          className={classNames('GoodsList', {
+            'GoodsList--hide': hidden,
```
mykola-yarchuk, line 75: same, use conditional rendering

src/components/GoodsList/GoodsList.js
```diff
@@ -0,0 +1,150 @@
...
+            onClick={this.reverseGoods}
+          >
+            Reverse
+          </button>
+          <button
+            type="button"
+            className="GoodsList__modifieButton"
+            onClick={() => {
+              this.setState({
+                sortBy: 'alphabet',
```
mykola-yarchuk, line 97: if you use a switch case, is better to create some constant like:
```
const SORT_CASES: {
  alphabet: 'alphabet',
  length: 'length',
}
```
and use them  here
```suggestion
                sortBy: SORT_CASES.alphabet,
```

src/components/GoodsList/GoodsList.js
```diff
@@ -0,0 +1,150 @@
...
+            <option>1</option>
+            <option>2</option>
+            <option>3</option>
+            <option>4</option>
+            <option>5</option>
+            <option>6</option>
+            <option>7</option>
+            <option>8</option>
+            <option>9</option>
+            <option>10</option>
```
mykola-yarchuk, line 140: use loop to create options

mykola-yarchuk (changes requested):

PR #433, opened 2021-05-24
Solution

igor-stasiv:
[DEMO LINK](https://igor-stasiv.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import './App.css';
 
+import { GoodsList } from './components/GoodsList';
+
 const goodsFromServer = [
   'Dumplings',
   'Carrot',
@@ -17,8 +19,6 @@ const goodsFromServer = [
 class App extends React.Component {
   state = {
     isGoodsShown: false,
-    isReversed: false,
-    lengthFilter: 1,
   }
 
   showGoods = () => {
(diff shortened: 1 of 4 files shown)


pkuzmenkoo (approved):
good decision, everything is clear

src/App.js
```diff
@@ -14,11 +14,129 @@ const goodsFromServer = [
...
+            Set filter by word length:
+            <select
+              value={lengthFilter}
+              onChange={({ target }) => {
+                this.setState({
+                  lengthFilter: target.value,
+                });
+              }}
+            >
+              <option>1</option>
```
varseniuk, line 96: Let's not write so many lines of code, but use an array of numbers from 1 to 10 and map `options` from it

varseniuk (changes requested):
Good code! 👍 
But let's separate the logic to different components (read my comment)

src/components/GoodsList.jsx
```diff
@@ -0,0 +1,125 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const numbers = [];
+
+for (let i = 1; i <= 10; i += 1) {
+  numbers.push(i);
```
varseniuk, line 7: You could use `new Array(10)` with `fill` method as a variant

varseniuk (approved):
Good job!

PR #435, opened 2021-05-24
Solution v0.1 without advanced task

SkyPKWolf:
[DEMO LINK](https://SkyPKWolf.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,7 @@
 import React from 'react';
+import classNames from 'classnames';
 import './App.css';
+import { GoodsList } from './components/GoodsList/GoodsList';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -14,11 +16,120 @@ const goodsFromServer = [
...
+                className="btn"
+                type="button"
+                onClick={this.reset}
+              >
+                Reset
+              </button>
+            </>
+          )}
+
+        {isVisible && <GoodsList goods={sortedGoods} />}
```
igor-stasiv, line 128: You can add this to the previous check

igor-stasiv (approved):
Nice solution!

src/App.js
```diff
@@ -14,11 +16,118 @@ const goodsFromServer = [
...
+
+  sortByLength = () => {
+    this.setState({
+      sortBy: 'length',
+    });
+  }
+
+  render() {
+    const { goods, isVisible, isReversed, sortBy } = this.state;
+    const sortedGoods = goods.filter(item => item);
```
varseniuk, line 60:
```suggestion
    const sortedGoods = [...goods]
```
There's a better way to make a copy of an array

src/App.js
```diff
@@ -14,11 +16,118 @@ const goodsFromServer = [
...
+
+    if (isReversed) {
+      sortedGoods.reverse();
+    }
+
+    return (
+      <div className="App">
+        <h1>Goods</h1>
+        {isVisible === false
+          ? (
```
varseniuk, line 81:
```suggestion
        {!isVisible ? (
```

varseniuk (changes requested):
The code is good, but let's split the logic between the components. Your `GoodsList` should only receive the goodsFromServer and all sorting logic should be done inside it (not in the `App`)

src/components/GoodsList/GoodsList.jsx
```diff
@@ -0,0 +1,127 @@
...
+      sortedGoods.reverse();
+    }
+
+    return (
+      <div className="App">
+        <h1>Goods</h1>
+        {!isVisible
+          ? (
+            <button
+              type="button"
```
varseniuk, line 69: This button and `isVisible` property should be on the upper level (in the `App`) and according to `isVisible` you should render the list:
`{isVisible && <GoodsList goods={goodsFromServer} />)`

varseniuk (approved):
Good! 
Pay attention to my comment

PR #437, opened 2021-05-25
solution

vitalya2626:
[DEMO LINK](https://vitalya2626.github.io/react_list-of-goods/)

### DIFF
--- a/src/ListOFGoods.js
+++ b/src/ListOFGoods.js
@@ -1,0 +1,88 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class ListOfGoods extends React.Component {
+  state = {
+    goods: [...this.props.goods],
+  }
+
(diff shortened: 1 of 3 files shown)


src/ListOFGoods.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class ListOfGoods extends React.Component {
+  state = {
+    goods: [...this.props.goods],
+  }
```
IvanGrekov, line 7: Why do you add goods to state? You can just use this.props for rendering list

IvanGrekov (approved):
Looks OK, but I left one remark. Pay attention to it in the future.

src/ListOFGoods.js
varseniuk: Good point

PR #441, opened 2021-05-25
Solution

TimurHrynenko:
- [DEMO LINK](https://TimurHrynenko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import './App.css';
+import { GoodsList } from './Components/GoodsList';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


src/Components/GoodsList.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const GoodsList = ({ preparedGoods }) => (
```
rauzok, line 4:
```suggestion
export const GoodsList = React.memo(
          ({ preparedGoods }) => (
```

rauzok (approved):
Awesome

src/App.js
```diff
@@ -14,11 +14,135 @@ const goodsFromServer = [
...
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    goods: goodsFromServer,
+    visibleGoods: [...goodsFromServer],
+    started: false,
+    productNameLength: [...Array(10).keys()].map(i => i + 1),
```
mykola-yarchuk, line 22: don't save that in state, it never changes

src/App.js
```diff
@@ -14,11 +14,135 @@ const goodsFromServer = [
...
 
-const App = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    goods: goodsFromServer,
```
mykola-yarchuk, line 19: don't save two arrays

mykola-yarchuk (changes requested):

src/App.js
```diff
@@ -14,11 +14,135 @@ const goodsFromServer = [
...
+
+  startHandler = () => {
+    this.setState(state => ({
+      started: !state.started,
+    }));
+  }
+
+  reverseHandler = () => {
+    this.setState(state => ({
+      visibleGoods: state.visibleGoods.reverse(),
```
mykola-yarchuk, line 34: don't mutate the previous state, use a copy

src/App.js
TimurHrynenko: Actually I don't understand how the length filter should work then. We always need the original array, otherwise, once we remove short words, we will not be able to return them in any way except reset.

VinnieJ-2k20: You should always work with available list `goodsFromServer` - prepare it in `App` and pass it to `GoodsList`.
Just remember not to mutate original array.

Hint - you can save sort and filter conditions in state.

VinnieJ-2k20 (changes requested):
Let's restructure your `App` a little bit.

VinnieJ-2k20 (approved):

PR #445, opened 2021-05-25
first solution

scarlettjohanssons:
[DEMO LINK](https://scarlettjohanssons.github.io/react_list-of-goods/

### DIFF
--- a/src/Goods.jsx
+++ b/src/Goods.jsx
@@ -42,7 +42,9 @@ export class Goods extends React.Component {
 
   render() {
     const { isReversed, isSorted, goodLength } = this.state;
-    const goodsList = this.props.goods.filter(good => good.length >= goodLength);
+    const goodsList = this.props.goods.filter(
+      good => good.length >= goodLength,
+    );
 
     goodsList.sort((first, second) => {
       switch (isSorted) {


rauzok (approved):
Nice!

src/App.js
```diff
@@ -14,11 +15,48 @@ const goodsFromServer = [
...
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    goodsList: [...goodsFromServer],
+    show: false,
+    showBtn: true,
```
mykola-yarchuk, line 22: don't save goods in the state
use only one boolean for rendering button and goods

src/Goods.jsx
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class Goods extends React.Component {
+  state = {
+    goods: [...this.props.goods],
```
mykola-yarchuk, line 6: don't save props in the state, just use goods from the props

mykola-yarchuk (changes requested):

src/Goods.jsx
```diff
@@ -0,0 +1,118 @@
...
+        </button>
+        <select
+          value={goodLength}
+          onChange={this.changeList}
+        >
+          {Array(10).fill(0).map((x, i) => (
+            <option
+              key={Math.floor(Math.random() * 100000)}
+              value={i + 1}
+            >
```
mykola-yarchuk, line 99: is better to create an array in advance, right now you create an array for every render, move this code out of render method

mykola-yarchuk (approved):

PR #448, opened 2021-05-25
solution

RuslanaPodshyvalova:
[DEMO LINK](https://RuslanaPodshyvalova.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import { GoodsList } from './components/GoodsList';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -14,11 +16,107 @@ const goodsFromServer = [
...
+      }
+    });
+
+    if (isReversed) {
+      goodsByCriteria.reverse();
+    }
+
+    return (
+      <div className="App">
+        <button
```
varseniuk, line 72:
```suggestion
        {isStartScript && <button
```
Better to use conditional rendering instead of setting `display: none`

src/App.js
```diff
@@ -14,11 +16,107 @@ const goodsFromServer = [
...
+          className={classname('start', {
+            show: isStartScript,
+            hide: !isStartScript,
+          })}
+          onClick={this.startSript}
+        >
+          Start
+        </button>
+        <div
+          className={classname({ hide: isStartScript })}
```
varseniuk, line 83:
```suggestion
        {!isStartScript && <div
```

varseniuk (approved):
Logic is good, but it would be great to move all the sortings logic inside the `GoodsList` component.

PR #451, opened 2021-05-25
Solution

SergeyZub:
[DEMO LINK](https://SergeyZub.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import GoodsList from './GoodsList';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


rauzok (changes requested):
Nice work!
But **reverse button** does not work correct, it needs to reverse current state.
Your **reverse button** reverse initial state.

rauzok (approved):

src/App.js
```diff
@@ -14,11 +15,85 @@ const goodsFromServer = [
...
+            onClick={this.sortByLengthList}
+          >
+            Sort by length
+          </button>
+
+          <button
+            type="button"
+            onClick={() => {
+              this.setState({
+                arrayList: [...goods],
```
varseniuk, line 84:
```suggestion
                arrayList: [...goodsFromServer],
```
No need to create identical variable `goods`

src/App.js
```diff
@@ -14,11 +15,85 @@ const goodsFromServer = [
...
 
-const App = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    goods: [...goodsFromServer],
```
varseniuk, line 20:
```suggestion
```
This value is redundant. No need to create the same variables

src/App.js
```diff
@@ -14,11 +15,85 @@ const goodsFromServer = [
...
+    displayList: false,
+    arrayList: [...goodsFromServer],
+  };
+
+  reverseList = () => {
+    this.setState(state => ({
```
varseniuk, line 26: Let's move all the sortings inside the `GoodsList` component.
It's good to separate the logic. `App` component should only have the button 'start' which will toggle `displayList`.
And according to the `displayList` you should render `GoodsList` component. There you should pass initial goods and sort them inside that component (not inside the `App`)

src/App.js
```diff
@@ -14,11 +15,85 @@ const goodsFromServer = [
...
+                arrayList: [...goods],
+              });
+            }}
+          >
+            Reset
+          </button>
+        </>
+        )}
+
+        {displayList && <GoodsList goods={arrayList} />}
```
varseniuk, line 93:
```suggestion
        {displayList && <GoodsList goods={goodsFromServer} />}
```
Pass initial goods an sort them there

varseniuk (changes requested):
The code is good, but let's split the logic

varseniuk (approved):
Good 👍

PR #452, opened 2021-05-25
 react_list-of-goods

Bogdan-Kotsupey:
[DEMO LINK](https://Bogdan-Kotsupey.github.io/react_list-of-goods/)

### DIFF
--- a/src/components/GoodsList.js
+++ b/src/components/GoodsList.js
@@ -1,0 +1,97 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class GoodsList extends React.Component {
+  state = {
+    goods: this.props.goods,
+    isReversed: false,
+    sortedBy: '',
+  }
+
+  reverse = () => {
+    this.setState(state => ({ isReversed: !state.isReversed }));
+  }
+
+  sortAlphabetically = () => {
+    this.setState(state => ({
+      ...state,
+      sortedBy: 'alphabetically',
+    }));
+  }
+
+  reset = () => {
+    this.setState({
+      isReversed: false,
+      sortedBy: '',
+    });
+  }
+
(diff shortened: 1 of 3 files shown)


rauzok (approved):
Good!

src/components/GoodsList.js
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class GoodsList extends React.Component {
+  state = {
+    goods: this.props.goods,
```
mykola-yarchuk, line 6: don't save props in state

mykola-yarchuk: just use goods from props

mykola-yarchuk (changes requested):

src/components/GoodsList.js
```diff
@@ -0,0 +1,108 @@
...
+    });
+  }
+
+  sortByLength = () => {
+    this.setState({ sortedBy: 'length' });
+  }
+
+  render() {
+    const { isReversed, sortedBy, length } = this.state;
+    const copy = [...this.props.goods].filter(
```
mykola-yarchuk, line 37: method filter don't mutate an array,  no need to copy it

src/components/GoodsList.js
```diff
@@ -0,0 +1,108 @@
...
+        >
+          Sort by length
+        </button>
+        <select
+          onChange={this.lengthFilter}
+          value={length}
+        >
+          {[...Array(11).keys()].map(element => (
+            <option value={element}>
+              {element}
```
mykola-yarchuk, line 95: is better to create an array in advance, right now you create new array for every render

mykola-yarchuk (changes requested):

PR #453, opened 2021-05-25
Solution-001

senkiv-oleh:
[DEMO LINK](https://senkiv-oleh.github.io/react_list-of-goods/)


src/App.js
```diff
@@ -14,11 +17,99 @@ const goodsFromServer = [
...
-  </div>
-);
+class App extends React.Component {
+  state = {
+    goods: [],
+  }
+
+  showGoodsList = () => {
+    this.setState({ goods: [...goodsFromServer] });
+  }
```
varseniuk, line 27: It would be good to change the logic here a bit. It would be better to create in the state the value, for example, `areGoodsVisible`, and render 'show' button and `GoodsList` according to this value:

`{this.state.areGoodsVisible && <GoodsList />}`
and
`{!this.state.areGoodsVisible && <button>}`
When button is clicked you should change that value:

```
showGoodsList = () => {
  this.setState(state => {(
      areGoodsVisible: !state.areGoodsVisible
    )}
  )
}

```

src/App.js
```diff
@@ -14,11 +17,99 @@ const goodsFromServer = [
...
+
+  render() {
+    const { goods } = this.state;
+
+    return (
+      <div className="App">
+        <h1>Goods</h1>
+        <button
+          type="button"
+          className={classNames('button', 'button--start', {
```
varseniuk, line 59: Let's use conditional rendering instead of setting `display:none`

src/App.js
```diff
@@ -14,11 +17,99 @@ const goodsFromServer = [
...
+
+  reverse = () => {
+    this.setState(state => ({
+      goods: [...state.goods].reverse(),
+    }));
+  }
+
+  sortAlphabetically = () => {
+    this.setState(state => ({
+      goods: [...state.goods].sort((a, b) => a.localeCompare(b)),
```
varseniuk, line 37: Also, it would be great to move all the sortings logic inside the `GoodsList` component.

varseniuk (approved):
Left some comments on how it may be improved

PR #457, opened 2021-05-25
solution without styles

dmitry-puhliakov:
[DEMO LINK](https://dmitry-puhliakov.github.io/react_list-of-goods/)


senkiv-oleh (approved):
Good

varseniuk (approved):
Good job!
But it would be great to move all the sortings inside the `GoodsList` component.

PR #459, opened 2021-05-26
Created sort list

YuriiLiso:
[DEMO LINK](https://yuriiliso.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,7 @@
+import classNames from 'classnames';
 import React from 'react';
 import './App.css';
+// import Page from './Page';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


Dima-Batiuk (approved):
Well done!)

src/App.js
```diff
@@ -14,11 +16,124 @@ const goodsFromServer = [
...
+                <li key={el}>
+                  {el}
+                </li>
+              ))}
+            </ul>
+
+            <div className="buttons">
+              <button
+                type="button"
+                className={classNames(isReversed ? 'active' : '')}
```
varseniuk, line 102:
```suggestion
                className={classNames({'active' : isReversed})}
```
Use classnames library properly

src/App.js
```diff
@@ -14,11 +16,124 @@ const goodsFromServer = [
...
+                onClick={this.reverseMethod}
+              >
+                Reverse
+              </button>
+
+              <button
+                type="button"
+                onClick={() => this.sortByField('byAlphabet')}
+                className={classNames(sortBy === 'byAlphabet' ? 'active' : '')}
```
varseniuk, line 111: Same here

src/App.js
```diff
@@ -14,11 +16,124 @@ const goodsFromServer = [
...
+              >
+                Sort alphabetically
+              </button>
+
+              <button
+                type="button"
+                onClick={() => this.sortByField('byLength')}
+                className={classNames(sortBy === 'byLength' ? 'active' : '')}
```
varseniuk, line 119: And here

src/App.js
```diff
@@ -14,11 +16,124 @@ const goodsFromServer = [
...
+
+  buttonSatus = () => {
+    this.setState({
+      isHidden: false,
+    });
+  }
+
+  reverseMethod = () => {
+    this.setState(state => ({
+      isReversed: !state.isReversed,
```
varseniuk, line 35: It would be great to move all the sortings inside the `GoodsList` component.

varseniuk (approved):
Code is good, but it would be better to split the logic between the components

PR #464, opened 2021-05-26
react list

SerhiiKolesnyk:
[DEMO LINK](https://SerhiiKolesnyk.github.io/react_list-of-goods/)

### DIFF
--- a/src/Components/Goods.jsx
+++ b/src/Components/Goods.jsx
@@ -1,0 +1,99 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class Goods extends React.Component {
+    state = {
+      isReversed: false,
+      sortMethod: null,
+      length: 1,
+    };
+
+    componentDidMount() {
+      this.defaultConditions = this.state;
+    }
+
+    reverse = () => {
+      this.setState(({ isReversed }) => ({ isReversed: !isReversed }));
+    }
+
+    sortFunction = (event) => {
+      const { name } = event.target;
+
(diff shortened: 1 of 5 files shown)


src/Components/Goods.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+class Goods extends React.Component {
+    state = {
+      isReversed: false,
+      sortMethod: null,
+      length: 1,
+    };
+
+    componentDidMount() {
+      this.defaultConditions = this.state;
```
mykola-yarchuk, line 12: copy object if you want to use it like that

mykola-yarchuk (changes requested):

src/Components/Goods.jsx
```diff
@@ -0,0 +1,89 @@
...
+      const visiableGoods = [...goods];
+
+      switch (sortMethod) {
+        case('sortByAlphabet'):
+        visiableGoods.sort((a, b) => a.localeCompare(b));
+        break;
+        case('sortByLength'):
+        visiableGoods.sort((a, b) => a.length - b.length);
+        break;
+        case('reverse'):
```
mykola-yarchuk, line 37: I believe that there are two ways of modifying goods, one it's sorting it and another one it's reversing, so it's should be a different value in the state for `reverse`.

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):

PR #467, opened 2021-05-26
solution

MinchukO:
[DEMO LINK](https://minchuko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -14,11 +14,90 @@ const goodsFromServer = [
...
+    goods: [...state.goods].sort((a, b) => a.length - b.length),
+  }));
+};
+
+render() {
+  return (
+    <div className="app">
+      <h1>Goods</h1>
+      <button
+        className={this.state.showList && 'visibility'}
```
varseniuk, line 58:
```suggestion
      {!showList && <button
```
Use conditional rendering instead of setting `display:block`

src/App.js
```diff
@@ -14,11 +14,90 @@ const goodsFromServer = [
...
+        type="button"
+        onClick={this.toggleListHandler}
+      >
+        Start
+      </button>
+      <button
```
varseniuk, line 64:
```suggestion
      {showList && <button
```

src/App.js
```diff
@@ -14,11 +14,90 @@ const goodsFromServer = [
...
+        className={!this.state.showList && 'visibility'}
+        type="button"
+        onClick={this.sortByLength}
+      >
+        Sort by length
+      </button>
+      <ul>
+        {this.state.showList && this.state.goods.map(product => (
+          <li key={product}>{product}</li>
```
varseniuk, line 94: Let's use a component approach and create a separate component `GoodsList`.

Also, it would be great to move all the sortings logic inside the `GoodsList` component.

varseniuk (changes requested):
Let's use conditional rendering

varseniuk (approved):

PR #470, opened 2021-05-26
solution

tanyakovchuk:
[DEMO LINK](https://taya294.github.io/react_list-of-goods/)


BilliZane (approved):
Looks good for me :+1:

src/App.jsx
```diff
@@ -0,0 +1,115 @@
...
+class App extends React.Component {
+  state = {
+    goods: [],
+  };
+
+  showGoods = () => {
+    this.setState({
+      goods: [...goodsFromServer],
+    });
+  }
```
varseniuk, line 28: Let's change logic here a bit. It would be better to create in the state the value, for example, `areGoodsVisible`, and render 'show' button and `GoodsList` according to this value:

`{this.state.areGoodsVisible && <GoodsList />}`
and
`{!this.state.areGoodsVisible && <button>}`
When button is clicked you should change that value:

```
showGoods = () => {
  this.setState(state => {(
      areGoodsVisible: !state.areGoodsVisible
    )}
  )
}

```

src/App.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+  sortAlphabetically = () => {
+    this.setState(state => ({
+      goods: [...state.goods].sort((a, b) => a.localeCompare(b)),
+    }));
+  }
+
+  sortByLength = () => {
+    this.setState(state => ({
```
varseniuk, line 43: It would be great to move all the sortings inside the `GoodsList` component.
It's good practice to split the logic. `App` component should only have the button 'start' which will toggle `areGoodsVisible`. And according to the `areGoodsVisible` should render `GoodsList` component. There you should pass initial goods and sort them inside that component (not inside the `App`)

varseniuk (changes requested):
Code is good, but let's use a component approach in react and split the logic

varseniuk (approved):

PR #483, opened 2021-05-26
solution

Namart:
[DEMO LINK](https://Namart.github.io/react_list-of-goods/)

### DIFF
--- a/src/components/GoodsList/GoodsList.jsx
+++ b/src/components/GoodsList/GoodsList.jsx
@@ -1,0 +1,91 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class GoodsList extends React.Component {
+  state = {
+    isReversed: false,
+    sortBy: '',
+  }
+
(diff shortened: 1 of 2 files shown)


src/components/GoodsList/GoodsList.jsx
```diff
@@ -0,0 +1,91 @@
...
+    this.setState({
+      isReversed: false,
+      sortBy: '',
+    });
+  }
+
+  render() {
+    const { goods } = this.props;
+
+    goods.sort((a, b) => {
```
varseniuk, line 26:
```suggestion
    const { goods } = this.props;
    
    const filteredGoods = [...goods]

    filteredGoods.sort((a, b) => {
```
Make a copy of the incoming array from props and do everything with it

varseniuk (changes requested):
Logic is good, but let's not change the initial array but create a copy of it

varseniuk (approved):

PR #485, opened 2021-05-26
Develop

vova-myronenko:
[DEMO LINK](https://vova-myronenko.github.io/react_list-of-goods/)


varseniuk (approved):
It would be great to move all the sortings logic inside the `GoodsList` component.

vova-myronenko:
> It would be great to move all the sortings logic inside the `GoodsList` component.
> It's good to split the logic. `App` component should only have the button 'start' which will toggle the render of the `GoodsList`. There you should pass initial goods and sort them inside that component (not inside the `App`)

Ok

PR #554, opened 2021-09-06
Develop

alex-glotov67:
- [DEMO LINK](https://alex-glotov67.github.io/react_list-of-goods/)

### DIFF
--- a/src/components/GoodsList/GoodsList.tsx
+++ b/src/components/GoodsList/GoodsList.tsx
@@ -44,7 +44,7 @@ export class GoodsList extends React.PureComponent<Props, State> {
     });
   };
 
-  handleChange = (event: any) => {
+  handleChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
     const { value } = event.target;
 
     this.setState({ minLength: value });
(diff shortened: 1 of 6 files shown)


Moroz-Dmytro (approved):
Good job 👍  But only your select displays goods with MIN selected length, not MAX

alex-glotov67:
> Good job 👍 But only your select displays goods with MIN selected length, not MAX

But in task `All the previous buttons
  should now show only goods having length >= than the selected value.` So if I choose the length of 2 it means that I must show all goods with lengths that are equal or bigger that we choose. So I understand it like this)

src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,163 @@
...
+  sortBy: string;
+  allLengths: string[];
+  minLength: string;
+};
+
+export class GoodsList extends React.PureComponent<Props, State> {
+  state: State = {
+    isReversed: false,
+    sortBy: 'default',
+    allLengths: ['1', '2', '3', '5', '6', '7', '8', '9', '10'],
```
ArtemTopolskyi, line 20: Это поле не отображает состояние приложение, вынеси его из state

ArtemTopolskyi (changes requested):

ArtemTopolskyi (approved):

PR #578, opened 2021-09-10
Solution

Vova-Stoliar:
[DEMO LINK](https://Vova-Stoliar.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -43,18 +43,24 @@ export class App extends React.PureComponent<{}, State> {
         case 'Sort by length':
           return {
             ...currentState,
+            sortGoodsByLength: !currentState.sortGoodsByLength,
+            reset: true,
+            sortGoodsByLetter: true,
+            reverseGoods: true,
             changedGoods: [...currentState.changedGoods].sort(
               (a, b) => a.length - b.length,
             ),
-            sortGoodsByLength: !currentState.sortGoodsByLength,
           };
         case 'Sort alphabetically':
           return {
             ...currentState,
+            sortGoodsByLetter: !currentState.sortGoodsByLetter,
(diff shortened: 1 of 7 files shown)


set1one (approved):
Great!

src/App.tsx
```diff
@@ -1,5 +1,8 @@
+/* eslint-disable no-console */
```
danheim, line 1: You are not allowed to do this

src/App.tsx
```diff
@@ -14,11 +17,167 @@ const goodsFromServer: string[] = [
...
+      reset,
+      reverseGoods,
+      maxWordLength,
+    } = this.state;
+
+    let filterGoods = changedGoods;
+
+    if (+maxWordLength >= 0) {
+      filterGoods = changedGoods.filter(good => good.length <= +maxWordLength);
+    }
```
danheim, line 122: avoid filtering directly in render

src/button/button.tsx
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable no-console */
```
danheim, line 1: again

src/select/Select.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const Select: React.FC<Props> = React.memo((props) => {
+  const { wordsLength } = props;
+
+  return (
+    <>
+      <option value="Word length" disabled>
+        Word length
+      </option>
+      {wordsLength.map(wordLength => {
+        return (
```
danheim, line 16: avoid return

danheim (changes requested):
good start

danheim (approved):

PR #590, opened 2021-09-12
Solution

Clonvhs:
[DEMO LINK](https://clonvhs.github.io/react_person/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
+import { GoodsList } from './components/GoodsList';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -14,11 +15,45 @@ const goodsFromServer: string[] = [
...
-);
+type State = {
+  goodsFromServer: string[];
+  showList:boolean;
+};
+
+class App extends React.Component {
+  state:State = {
+    goodsFromServer,
+    showList: false,
```
alex-glotov67, line 26: variable with boolean types better named like `isListVisible` - it nice when it starts with `is`

alex-glotov67 (approved):

src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,123 @@
...
+          <option value="1">1</option>
+          <option value="2">2</option>
+          <option value="3">3</option>
+          <option value="4">4</option>
+          <option value="5">5</option>
+          <option value="6">6</option>
+          <option value="7">7</option>
+          <option value="8">8</option>
+          <option value="9">9</option>
+          <option value="10">10</option>
```
Elena-brv, line 104: what if it will be 1000 options? better use cycle or array method for such cases

Elena-brv (approved):

PR #600, opened 2021-10-12
Solution

Lenur29:
https://Lenur29.github.io/react_list-of-goods/


src/App.tsx
```diff
@@ -14,11 +15,119 @@ const goodsFromServer: string[] = [
...
+
+  render() {
+    const {
+      isButtonClicked,
+      goods,
+      isReversed,
+      sortBy,
+    } = this.state;
+
+    const visibleGoods = goods.slice();
```
vkryvytskyy, line 66:
```suggestion
    const visibleGoods = [...goods];
```

PR #614, opened 2021-10-18
solution

VladSaen:
- [DEMO LINK](https://VladSaen.github.io/react_list-of-goods/)


src/App.tsx
```diff
@@ -14,11 +14,129 @@ const goodsFromServer: string[] = [
...
+
+  clearSort = () => {
+    this.setState({ sortType: 'none' });
+  };
+
+  render() {
+    const {
+      goods, isVisible, isReverse, sortType,
+    } = this.state;
+    const goodsCopy = [...goods].reverse();
```
aliceblunt8, line 56: why do you do reverse by default?

src/App.tsx
```diff
@@ -14,11 +14,129 @@ const goodsFromServer: string[] = [
...
+                  onClick={this.sortByLength}
+                >
+                  name length
+                </button>
+
+                <button
+                  type="button"
+                  onClick={this.clearSort}
+                >
+                  none
```
aliceblunt8, line 118: this button doesn't work like reset

aliceblunt8 (changes requested):

src/App.tsx
VladSaen: oops

src/App.tsx
aliceblunt8: it works only after sort, but it should also reset reverse ;)

PR #630, opened 2021-10-18
solution

Saharij:
[DEMO](https://Saharij.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -20,15 +20,21 @@ type State = {
   isReverse: boolean;
   isSortedByAlph: boolean;
   isSortedByLeng: boolean;
+  showList: boolean;
   selectNumber: number;
 };
 
 class App extends React.Component<{}, State> {
-  state = {
+  state: State = {
     isReverse: false,
     isSortedByAlph: false,
     isSortedByLeng: false,
+    showList: false,
     selectNumber: 1,
+  };
+
+  handleStart = () => {
+    this.setState({ showList: true });
   };
 
   handleButtonReverse = () => {
@@ -65,6 +71,7 @@ class App extends React.Component<{}, State> {
       isReverse,
       isSortedByAlph,
       isSortedByLeng,
+      showList,
       selectNumber,
     } = this.state;
     const goods = [...goodsFromServer].filter(
@@ -73,10 +80,8 @@ class App extends React.Component<{}, State> {
     const arrForSelect = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
     const buttonClass = 'button is-primary';
 
-    goods.filter(item => item.length >= selectNumber);
-
     if (isSortedByAlph) {
-      goods.sort((g1, g2) => g1.localeCompare(g2));
+      goods.sort();
     }
 
     if (isSortedByLeng) {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -14,11 +16,132 @@ const goodsFromServer: string[] = [
...
-);
+type State = {
+  isReverse: boolean;
+  isSortedByAlph: boolean;
+  isSortedByLeng: boolean;
+  selectNumber: number;
+};
+
+class App extends React.Component<{}, State> {
+  state = {
```
aliceblunt8, line 27:
```suggestion
  state: State = {
```

src/App.tsx
```diff
@@ -14,11 +16,132 @@ const goodsFromServer: string[] = [
...
+      isSortedByLeng,
+      selectNumber,
+    } = this.state;
+    const goods = [...goodsFromServer].filter(
+      item => item.length >= selectNumber,
+    );
+    const arrForSelect = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+    const buttonClass = 'button is-primary';
+
+    goods.filter(item => item.length >= selectNumber);
```
aliceblunt8, line 76: you do filter twice, furthermore filter doesn't mutate array in place, so second filter does nothing at all

src/App.tsx
```diff
@@ -14,11 +16,132 @@ const goodsFromServer: string[] = [
...
+      goods.sort((g1, g2) => g1.localeCompare(g2));
+    }
+
+    if (isSortedByLeng) {
+      goods.sort((g1, g2) => g1.length - g2.length);
+    }
+
+    if (isReverse) {
+      goods.reverse();
+    }
```
aliceblunt8, line 88: maybe add sortBy to the state with different values and use switch case for choosing which sort to use

src/App.tsx
```diff
@@ -14,11 +16,132 @@ const goodsFromServer: string[] = [
...
+    const goods = [...goodsFromServer].filter(
+      item => item.length >= selectNumber,
+    );
+    const arrForSelect = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+    const buttonClass = 'button is-primary';
+
+    goods.filter(item => item.length >= selectNumber);
+
+    if (isSortedByAlph) {
+      goods.sort((g1, g2) => g1.localeCompare(g2));
```
aliceblunt8, line 79: this will work the same

```suggestion
      goods.sort();
```

src/App.tsx
```diff
@@ -14,11 +16,132 @@ const goodsFromServer: string[] = [
...
+    return (
+      <div className="App">
+        <h1 className="title is-1 mt-5">Goods</h1>
+
+        <div className="buttons__container">
+          <button
+            type="button"
+            className={this.state.isReverse
+              ? `${buttonClass} is-active`
+              : `${buttonClass}`}
```
aliceblunt8, line 99: you can use classnames lib for conditional classes
https://www.npmjs.com/package/classnames

src/components/GoodList/GoodList.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+type State = {
+  showList: boolean;
+};
+
+export class GoodList extends React.Component<Props, State> {
+  state = {
+    showList: false,
+  };
```
aliceblunt8, line 15: remove state showList state and start button from GoodList component and add it to App
so you can conditionally render GoodList
`showList && <GoodList />`

aliceblunt8 (changes requested):
GJ!

src/App.tsx
aliceblunt8: it was just recommendation ;) and in my opinion, proposed solution would look neater, but your solution is correct too

PR #636, opened 2021-10-19
Solution

DY-JS:
[DEMO LINK](https://DY-JS.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -14,11 +14,125 @@ const goodsFromServer: string[] = [
...
-);
+type State = {
+  goods: string[];
+  isReversed: boolean;
+  sortBy: string | null;
+  isRun: boolean;
+};
+
+class App extends React.Component<{}, State> {
+  state = {
```
aliceblunt8, line 25:
```suggestion
  state: State = {
```

src/App.tsx
```diff
@@ -14,11 +14,125 @@ const goodsFromServer: string[] = [
...
+          onClick={() => {
+            if (!isRun) {
+              this.start();
+            } else {
+              this.reset();
+            }
+          }}
+        >
+          {!isRun ? 'Start' : 'Reset'}
+        </button>
```
aliceblunt8, line 106: I think it's better to create separate buttons for each case than use condition inside onClick

src/App.tsx
```diff
@@ -14,11 +14,125 @@ const goodsFromServer: string[] = [
...
+
+        <div className="app__buttons">
+          <input
```
aliceblunt8, line 109: why do you use input instead buttons?

PR #659, opened 2021-11-12
Develop

Olexandr6:
[DEMO LINK](https://Olexandr6.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,13 @@
 import React from 'react';
+import { uuid } from 'uuidv4';
 import './App.css';
 
-const goodsFromServer: string[] = [
+interface Good {
+  key: string,
+  value: string
+}
+
+const goodsFromServer: Good[] = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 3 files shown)


DammneD (approved):
It's work, but i think you need to mark button  when sort or reverse method are active.

src/App.tsx
```diff
@@ -12,13 +15,151 @@ const goodsFromServer: string[] = [
...
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+interface State {
+  isStarted: boolean,
+  isReversed: boolean,
+  isDafaultOrder: boolean,
```
solaryasha, line 26:
```suggestion
  isDefaultOrder: boolean,
```

solaryasha (approved):
Nice !

Olexandr6:
> It's work, but i think you need to mark button when sort or reverse method are active.

Agree, it would be better)

PR #682, opened 2021-12-12
List of goods

lera-permiakova:
[DEMO LINK](https://lera-permiakova.github.io/react_list-of-goods/)


src/App.scss
```diff
@@ -0,0 +1,125 @@
...
+      top: 0;
+      left: 0;
+      z-index:-1;
+    }
+
+  &__container {
+    height: 100%;
+    padding: 40px;
+    background-color: rgba(255, 255, 255, 0.3);
+    // filter: brightness(0.94);
```
msaikun, line 43: Delete comments

msaikun (approved):
Looks great! But when I sort by length, the length of the ice cream is 9 letters, although it should be 8.
image

src/App.tsx
```diff
@@ -14,11 +14,148 @@ const goodsFromServer: string[] = [
...
+  };
+
+  sortByLength = () => {
+    this.setState({ sortBy: 'length' });
+  };
+
+  render() {
+    const {
+      count, goods, isReversed, sortBy, lengths, selectedNumb,
+    } = this.state;
```
ihor-jpeg, line 67:
```suggestion
    const {
      count,
      goods,
      isReversed,
      sortBy,
      lengths,
      selectedNumb,
    } = this.state;
```

src/App.tsx
```diff
@@ -14,11 +14,148 @@ const goodsFromServer: string[] = [
...
+              >
+                Reset
+              </button>
+              <button
+                className={count === 1 ? 'App__button' : 'disabled'}
+                type="button"
+                onClick={this.sortByLength}
+              >
+                Sort by length
+              </button>
```
ihor-jpeg, line 137: you do not really need to create a separate CSS class for it. you can write count === 1 to a variable and then conditionally render the needed buttons

ihor-jpeg (approved):
Well done! But when I press the reset button, everything resets apart from the word length selection. the number does not change back to 1 on the reset

PR #683, opened 2021-12-13
list of goods

Vitaliy-0:
[demo](https://vitaliy-0.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
+import { GoodList } from './components/GoodList';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -14,11 +15,43 @@ const goodsFromServer: string[] = [
...
-  </div>
-);
+interface State {
+  goods: string[],
+  isShown: boolean;
+}
+
+class App extends React.Component<{}, State> {
+  state = {
+    goods: goodsFromServer,
```
ihor-jpeg, line 25: it's best to just make a copy straight away to avoid mutation in the future

src/components/GoodList.tsx
```diff
@@ -0,0 +1,153 @@
...
+    return (
+      <div className="goods">
+        <button
+          className={classNames('btn',
+            { 'btn--selected': isReversed })}
+          type="button"
+          onClick={this.reverse}
+        >
+          Reversed
+          {`(${isReversed})`}
```
ihor-jpeg, line 90: i think this might be not needed

ihor-jpeg (approved):
Well done!

PR #687, opened 2021-12-13
Solution

dariia-romanova:
[DEMO LINK](https://dariia-romanova.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
+import { ListOfGoods } from './components/ListOfGoods';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -14,11 +15,121 @@ const goodsFromServer: string[] = [
...
+  };
+
+  reset = () => {
+    this.setState({ sortBy: '', isReversed: false });
+  };
+
+  render() {
+    const {
+      goods, isVisible, isReversed, sortBy,
+    } = this.state;
```
ihor-jpeg, line 58:
```suggestion
    const {
      goods,
      isVisible,
      isReversed,
      sortBy,
    } = this.state;
```

src/components/ListOfGoods/ListOfGoods.tsx
```diff
@@ -0,0 +1,17 @@
...
+type Props = {
+  goods: string[],
+};
+
+export const ListOfGoods: React.FC<Props> = ({ goods }) => {
+  return (
+    <ul className="list">
+      {
+        goods.map(good => (
+          <li key={good} className="list__item">{good}</li>
```
ihor-jpeg, line 12: map's callback also accepts the index. you can use that as a stronger version of the key.

ihor-jpeg (approved):
Well done!

src/components/ListOfGoods/ListOfGoods.tsx
dariia-romanova: I thought index can't be used as a key. I get an error when I use it.

ihor-jpeg: Yes, you are right. I thought of an id for some reason. But the items in the array are just strings, so they don't have IDs. Never mind the previous comment.

PR #689, opened 2021-12-13
Solution of list-of-goods

Dima-Ziakhor:
[DEMO LINK](https://Dima-Ziakhor.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
 import React from 'react';
-import './App.css';
+import { ListOfGoods } from './components/ListOfGoods';
+
+import './App.scss';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -14,11 +16,110 @@ const goodsFromServer: string[] = [
...
+  start = () => (
+    this.setState({
+      render: true,
+    }));
+
+  reverse = () => {
+    this.setState((state) => ({
+      goods: state.goods.reverse(),
+    }));
+  };
```
dariia-romanova, line 39: I think it's better not to mutate goods array in state. Same for sorting

dariia-romanova (approved):
Good job)

src/components/ListOfGoods.tsx
```diff
@@ -0,0 +1,21 @@
...
+    <ul className="listOfGoods">
+      {
+        goods.map(item => (
+          <li key={item} className="listOfGoods__item">
+            {item}
+          </li>
+        ))
+      }
+    </ul>
+  </>
```
ihor-jpeg, line 20: why use fragment here?

src/App.tsx
```diff
@@ -14,11 +16,108 @@ const goodsFromServer: string[] = [
...
+    this.setState((state) => ({
+      goods: [...state.goods].reverse(),
+    }));
+  };
+
+  sortByLetters = () => {
+    this.setState((state) => ({
+      goods: [...state.goods].sort((a, b) => a.localeCompare(b)),
+    }));
+  };
```
ihor-jpeg, line 45: when you sort the goods you don't really care about their initial state

src/App.tsx
```diff
@@ -14,11 +16,108 @@ const goodsFromServer: string[] = [
...
+    this.setState({
+      goods: [...goodsFromServer],
+    });
+  };
+
+  sortByLength = () => {
+    this.setState((state) => ({
+      goods: [...state.goods].sort((a, b) => a.length - b.length),
+    }));
+  };
```
ihor-jpeg, line 57: why do you need the previous state here?

ihor-jpeg (commented):
Well done! But let's work on a few fixes

ihor-jpeg (changes requested):
.

ihor-jpeg (approved):
well done!

PR #698, opened 2021-12-13
Solution

alinaShchavinska:
[DEMO LINK](https://alinaShchavinska.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -20,7 +20,6 @@ type State = {
   sortBy: string,
   isRendered: boolean,
   isReversed: boolean,
-  selectedLength: number,
 };
 
 class App extends React.PureComponent<{}, State> {
@@ -29,7 +28,6 @@ class App extends React.PureComponent<{}, State> {
     isRendered: false,
     isReversed: false,
     sortBy: '',
-    selectedLength: 1,
   };
 
   showList = () => (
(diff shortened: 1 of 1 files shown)


tmrsl (approved):
GJ and so pretty design 🔥

src/App.tsx
```diff
@@ -14,11 +15,126 @@ const goodsFromServer: string[] = [
...
+    this.setState({
+      isReversed: false,
+      selectedLength: 1,
+      sortBy: '',
+    })
+  );
+
+  render() {
+    const {
+      goods, isRendered, isReversed, sortBy, selectedLength: itemLength,
```
ihor-jpeg, line 63: are you sure you need selectedLength: itemLength ?

src/App.tsx
```diff
@@ -14,11 +15,126 @@ const goodsFromServer: string[] = [
...
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  goods: string[],
+  sortBy: string,
+  isRendered: boolean,
+  isReversed: boolean,
+  selectedLength: number,
```
ihor-jpeg, line 23: what is  selectedLength: number, used for?

src/App.tsx
```diff
@@ -14,11 +15,126 @@ const goodsFromServer: string[] = [
...
+
+  render() {
+    const {
+      goods, isRendered, isReversed, sortBy, selectedLength: itemLength,
+    } = this.state;
+
+    let copyGoods = [...goods].filter((good) => good.replace(' ', '').length >= itemLength);
+
+    if (sortBy) {
+      copyGoods = [...copyGoods].sort((goodA, goodB) => {
```
ihor-jpeg, line 69: why do you make a copy of a copy?

ihor-jpeg (changes requested):
Good work! But let's fix a few things

src/App.tsx
alinaShchavinska: Wanted to do advanced tasks with Select Button ( All the previous buttons
  should now show only goods having length >= than the selected value).  But it did not work. And I forgot to delete what I don't use

ihor-jpeg (approved):
well done!

PR #708, opened 2021-12-14
react_list-of-goods task

lanebx:
[list-of-goods](https://lanebx.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
+/* eslint-disable no-console */
 import React from 'react';
 import './App.css';
+import { GoodsList } from './components/GoodsList';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


Nadiyahr (approved):
Great!

src/App.tsx
```diff
@@ -14,11 +16,141 @@ const goodsFromServer: string[] = [
...
+                  onClick={() => {
+                    this.setState((state: State) => ({ showList: !state.showList }));
+                  }}
+                >
+                  Start
+                </button>
+              )
+              : (
+                <GoodsList
+                  props={goods.filter((_item, i) => i < +this.state.value)}
```
ihor-jpeg, line 80: make a variable for this one. it's quite difficult to understand what is going on here.
also, it filters by index. it should be filtering by word length

src/App.tsx
```diff
@@ -14,11 +16,141 @@ const goodsFromServer: string[] = [
...
+      }),
+    }));
+  };
+
+  render() {
+    const {
+      goods, showList,
+    } = this.state;
+
+    console.log(this.state.value);
```
ihor-jpeg, line 59: you don't need this

src/App.tsx
```diff
@@ -14,11 +16,141 @@ const goodsFromServer: string[] = [
...
+    this.setState((state: State) => ({
+      goods: [...state.goods].reverse(),
+    }));
+  };
+
+  sortAlphabetically = () => {
+    this.setState((state: State) => ({
+      goods: state.goods.sort((a: string, b: string) => {
+        return a.localeCompare(b);
+      }),
```
ihor-jpeg, line 42: don't make such complex constructs

src/App.tsx
```diff
@@ -14,11 +16,141 @@ const goodsFromServer: string[] = [
...
+  goods: string[];
+  showList: boolean;
+  value: string;
+}
+
+class App extends React.Component<{}, State> {
+  state: State = {
+    goods: [...goodsFromServer],
+    showList: false,
+    value: `${goodsFromServer.length}`,
```
ihor-jpeg, line 29: you made this value a string and then in the method on line 80 you convert it back to number. seems a bit counterproductive

ihor-jpeg (commented):
Good work. but make your code easier to read and understand. don't make too complex constructs. creating a variable and then using it is okay and maybe it will make your code longer, but a lot easier to read

ihor-jpeg (changes requested):
.

src/App.tsx
lanebx: Is it a word length filter?
Снимок экрана 2021-12-09 в 14 36 23

ihor-jpeg: yep. it's correct now!

ihor-jpeg (approved):
well done

PR #710, opened 2021-12-14
List of goods (sort, reverse)

WorkAccCom:
[DEMO LINK](https://workacccom.github.io/react_list-of-goods/)

### DIFF
--- a/src/components/GoodsList/GoodsList.tsx
+++ b/src/components/GoodsList/GoodsList.tsx
@@ -1,0 +1,126 @@
+import React from 'react';
+import './GoodsList.scss';
+import '../../styles/button.scss';
+
+interface Props {
+  goods: string[],
+}
+
+interface State {
+  goods: string[],
+  sortBy: string,
+}
+
+export class GoodsList extends React.Component<Props> {
+  state = {
+    goods: [...this.props.goods],
+    sortBy: 'alphab',
+  };
+
+  reverse = () => (
+    this.setState((state: State) => ({
+      goods: state.goods.reverse(),
+    }))
+  );
+
+  sortAlphab = () => {
+    this.setState({
+      sortBy: 'alphab',
+    });
+
+    this.sort();
+  };
+
+  length = () => {
+    this.setState({
+      sortBy: 'length',
+    });
+
+    this.sort();
+  };
+
+  reset = () => {
+    this.setState({
+      goods: [...this.props.goods],
+    });
+  };
+
+  sort = () => (
+    this.setState((state: State) => ({
+      goods: state.goods.sort(
+        (productA, productB) => {
+          switch (state.sortBy) {
+            case 'alphab': {
+              return productA.localeCompare(productB);
+            }
+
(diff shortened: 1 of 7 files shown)


src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  sortAlphab = () => {
+    this.setState({
+      sortBy: 'alphab',
+    });
+
+    this.sort();
+  };
+
+  length = () => {
```
TatianaMorozova04, line 34: I think "sortLength" method name would be better

TatianaMorozova04 (approved):
Good work!

src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,126 @@
...
+          >
+            Sort by length
+          </button>
+        </div>
+
+        <ul className="goods-list__list">
+          {goods.map((product, index) => (
+            <li
+              className="goods-list__list-item"
+              key={`${product} ${index * Math.random()}`}
```
ihor-jpeg, line 117: you can just use the product as a key

src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,126 @@
...
+interface Props {
+  goods: string[],
+}
+
+interface State {
+  goods: string[],
+  sortBy: string,
+}
+
+export class GoodsList extends React.Component<Props> {
```
ihor-jpeg, line 14: use your State interface here as well

src/components/GoodsList/GoodsList.tsx
ihor-jpeg: yes, function names should start with verbs to tell what they do

ihor-jpeg (changes requested):
Well done. But let's fix a few things

src/components/GoodsList/GoodsList.tsx
WorkAccCom: I'm advisedly used Math.random  here, because, what if in this array will arise a two equals products?

WorkAccCom: please, kindly, answer me am I right or not?)

src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,122 @@
...
+          >
+            Sort by length
+          </button>
+        </div>
+
+        <ul className="goods-list__list">
+          {goods.map((product, index) => (
+            <li
+              className="goods-list__list-item"
+              key={`${product} ${index * Math.random()}`}
```
ihor-jpeg, line 113: Math.random returns random values, but random is not the same as unique. The keys should be unique. Math.random can return the same value twice.

ihor-jpeg (changes requested):

ihor-jpeg (approved):
Well done!

PR #736, opened 2022-01-24
Solution

AlexAzalor:
- [DEMO LINK](https://AlexAzalor.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,11 @@
 import React from 'react';
-import './App.css';
+import classNames from 'classnames';
 
-const goodsFromServer: string[] = [
+import { GoodsList } from './GoodsList';
+
+import './App.scss';
+
+const goodsFromServer = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -14,11 +18,100 @@ const goodsFromServer: string[] = [
...
+    this.setState((state) => ({
+      goods: [...state.goods].sort(),
+    }));
+  };
+
+  sortByLength = () => {
+    this.setState((state) => ({
+      goods: [...state.goods].sort((a, b) => a.length - b.length),
+    }));
+  };
```
Rozdorozhnii, line 58: As I understood your handlers shouldn't change array, handlers should change some simple fields of state.
Better change copy of array in render method.

Rozdorozhnii (changes requested):

src/App.tsx
```diff
@@ -14,11 +18,125 @@ const goodsFromServer: string[] = [
...
+
+  reverse = () => {
+    this.setState((state) => ({
+      isReversed: !state.isReversed,
+    }));
+  };
+
+  sort = () => {
```
Rozdorozhnii, line 55: better to use more informative names, also use suffix ('Handler') in naming to show that is function proceed some event

Rozdorozhnii (approved):

src/App.tsx
```diff
@@ -14,11 +18,125 @@ const goodsFromServer: string[] = [
...
+      initialGoods.sort((a, b) => a.localeCompare(b));
+    }
+
+    if (sortBy === Sortby.Length) {
+      initialGoods.sort((a, b) => a.length - b.length);
+    }
+
+    if (isReversed) {
+      initialGoods.reverse();
+    }
```
Olexandr6, line 93: Create a separate function `getVisibleGoods` for this logic

Olexandr6 (changes requested):
Good solution, but one thing should be optimized

src/App.tsx
AlexAzalor: for this { buttonVisible: buttonVisible === true } ?

Olexandr6 (approved):
Good job 👍

PR #737, opened 2022-01-24
solution_v1.0

Rozdorozhnii:
-[DEMO LINK](https://Rozdorozhnii.github.io/react_list-of-goods/)

### DIFF
--- a/src/components/GoodsList/GoodsList.tsx
+++ b/src/components/GoodsList/GoodsList.tsx
@@ -1,0 +1,215 @@
+import React from 'react';
+
+enum SortBy {
+  name = 'name',
+  length = 'length',
+  default = '',
+}
+
+enum Length {
+  One = 1,
+  Two,
+  Three,
+  Four,
+  Five,
+  Six,
+  Seven,
+  Eight,
+  Nine,
+  Ten,
+}
+
+type Props = {
+  goods: string[];
+};
+
+type State = {
+  isListVisible: boolean;
+  goods: string[];
+  isReversed: boolean;
+  sortBy: string;
+  visibleLength: number;
+};
+
+export class GoodsList extends React.Component<Props, State> {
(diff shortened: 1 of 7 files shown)


AlexAzalor (approved):

Olexandr6 (changes requested):
Nicely done 🔥 

But I noticed a bug, when you click the button `reverse` and then `sort`,  it doesn't sort alphabetically

src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,215 @@
...
+          return a.length - b.length;
+
+        default:
+          return 0;
+      }
+    });
+
+    if (isReversed) {
+      copyGoods.reverse();
+    }
```
Olexandr6, line 106: Btw, create a separate function for this logic. The method `render`, must be as concise as possible.

Rozdorozhnii:
I'm not sure that understood right requested changes.
I changed logic of sort buttons, now you can click them more than one time and order of sorting will change to oposite direction.

src/components/GoodsList/GoodsList.tsx
Rozdorozhnii: Misha showed this a bit complicated solution to sort rendering array

Olexandr6: @Rozdorozhnii In the video, it's difficult to show a clear and optimized solution that would be clear to everyone, so he does not emphasize this

Olexandr6: So put this logic in a separate function

Olexandr6 (changes requested):
Check my replies to previous comment and fix it

Olexandr6 (approved):
Great job 👍 Well done 🔥

PR #810, opened 2022-03-29
Develop

krolich500grn:
[DEMO LINK](https://krolich500grn.github.io/react_list-of-goods/)
 
 адвансед можливо потім дороблю)


src/App.tsx
```diff
@@ -14,11 +17,113 @@ const goodsFromServer: string[] = [
...
-);
+type Props = {};
+type State = {
+  isStarted: boolean,
+  goodsList: string[],
+  visibleGoods: string[],
+  isReversed: boolean,
+};
+
+class App extends React.Component<Props, State> {
```
jstmpelowycz, line 28:
```suggestion
type State = {
  isStarted: boolean,
  goodsList: string[],
  visibleGoods: string[],
  isReversed: boolean,
};

class App extends React.Component<{}, State> {
```

jstmpelowycz: If `Props` has no actual purpose, there is no need to create one. Its generic is replaceable with `{}`

src/App.tsx
```diff
@@ -14,11 +17,113 @@ const goodsFromServer: string[] = [
...
+  state = {
+    isStarted: false,
+    goodsList: goodsFromServer,
+    visibleGoods: goodsFromServer,
```
jstmpelowycz, line 32: This one isn't incorrect, however, I would rather create state properties indicating the way I need to render my list. For instance, `isReversed` and its further usage: if the flag is up I reverse goods list, otherwise I skip reversing

jstmpelowycz: I've done it so, thus you could check my code and decide if you like or not

src/additional_styles/button.css
```diff
@@ -0,0 +1,23 @@
...
+  outline: none;
+  padding: 13px;
+  position: relative;
+  text-align: center;
+  text-decoration: none;
+  touch-action: manipulation;
+  user-select: none;
+  -webkit-user-select: none;
+  width: auto;
+}
```
jstmpelowycz, line 23: I would recommend splitting rules into small groups according to their purpose (e.g. positioning, formatting etc.) - that's gonna make it much more readable😃

src/components/GoodsItem/GoodsItem.tsx
```diff
@@ -0,0 +1,9 @@
+type Props = {
+  good: string,
+};
+
+export const GoodsItem: React.FC<Props> = ({ good }) => (
+  <li className="Goods__item">
+    {good}
+  </li>
+);
```
jstmpelowycz, line 9: That bad idea to take `li` away from `ul`, plus you've forgot to explicitly indicate item key. I'm 100% sure Vova's gonna give a reject for this, so I'm gonna request changes before it happens)

src/App.tsx
jstmpelowycz: By doing this away, you avoid at least repeating yourself
```
  goodsList: goodsFromServer,
  visibleGoods: goodsFromServer,
```

jstmpelowycz (changes requested):
A few meaningful aspects to consider. Good job though, for sure!

src/additional_styles/button.css
krolich500grn: it's just ready & copied block of code 😁, okay)

src/components/GoodsItem/GoodsItem.tsx
krolich500grn: the key is delivered by props here, from the GoodsList component

src/App.tsx
```diff
@@ -14,11 +17,112 @@ const goodsFromServer: string[] = [
...
+  isStarted: boolean,
+  goodsList: string[],
+  visibleGoods: string[],
+  isReversed: boolean,
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    isStarted: false,
+    goodsList: goodsFromServer,
```
solaryasha, line 30: goodsList точно не варто зберігати в стейті, goodsFromServer у тебе завжди будуть в доступі

src/components/GoodsItem/GoodsItem.tsx
```diff
@@ -0,0 +1,9 @@
+type Props = {
+  good: string,
+};
+
+export const GoodsItem: React.FC<Props> = ({ good }) => (
```
solaryasha, line 5: react.memo теж можеш тут використати

solaryasha (changes requested):

solaryasha (approved):

PR #862, opened 2022-05-11
solution w0 advanced tasks

ShevchyshynRoman:
[DEMO LINK](https://ShevchyshynRoman.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
+import { GoodsList } from './Components/GoodsList/GoodsList';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -14,11 +15,108 @@ const goodsFromServer: string[] = [
...
+  };
+
+  showGoodsList = () => {
+    this.setState({ isGoodsVisible: true });
+  };
+
+  reverseGoods = () => {
+    this.setState((state) => {
+      return {
+        goods: [...state.goods].reverse(),
```
OleksandrNi, line 36: reverseGoods = () => {
    this.setState((state) => ({
     goods: [...state.goods].reverse(),
    }));
  };

ShevchyshynRoman: I agree, but in my opinion it is so obvious that we are returning the object

AlinaMatuschak: Agree with @OleksandrNi . Will be less code

PR #871, opened 2022-05-12
solution without styles

andriyshvets:
[DEMO](https://andriyshvets.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,11 +14,106 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  goods: string[],
+  isVisible: boolean,
+  length: number,
+};
+
+class App extends React.Component<{}, State> {
+  state = {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -14,11 +14,106 @@ const goodsFromServer: string[] = [
...
+
+class App extends React.Component<{}, State> {
+  state = {
+    goods: [...goodsFromServer],
+    isVisible: false,
+    length: 1,
+  };
+
+  showGoods = () => {
+    this.setState({ isVisible: true });
```
Vakolyyk, line 31: Аfter pressing the button again, the list should be hidden

src/App.tsx
```diff
@@ -14,11 +14,106 @@ const goodsFromServer: string[] = [
...
+    }));
+  };
+
+  render() {
+    const { goods, isVisible, length } = this.state;
+    const visibleGoods = goods.filter(good => good.length >= length);
+
+    return isVisible
+      ? (
+        <div className="App">
```
Vakolyyk, line 68: Change the rendering condition so that the list can be hidden

Vakolyyk (changes requested):
In general ok

danylo-onishchuk (approved):

PR #877, opened 2022-05-12
dev

kurochka-vladyslav:
https://kurochka-vladyslav.github.io/react_list-of-goods/


maxxnikiforov (approved):
I think all good! If you will have time try to add "<select>" - it's not so difficult.

MaksimusCRS (approved):
image
🤔

PR #902, opened 2022-06-04
Develop

ZUBOV-ILLIA:
[DEMO LINK](https://ZUBOV-ILLIA.github.io/react_list-of-goods/)

The filter contains only those values that are present in the array of products.


vlad-demchuk (approved):

denys-cheporniuk (changes requested):
Screenshot 2022-06-06 at 21 27 15

reset should also work with select

denys-cheporniuk (approved):

PR #903, opened 2022-06-04
list of good

Andriyk0:
[DEMO LINK](https://Andriyk0.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import React from 'react';
 import './App.scss';
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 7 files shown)


EnjSadman (approved):

src/App.tsx
```diff
@@ -14,11 +15,136 @@ const goodsFromServer: string[] = [
...
+    selectValue: 1,
+  };
+
+  showListOFGoods = () => {
+    this.setState({ visibleGoods: true });
+  };
+
+  reverseList = () => {
+    this.setState((state) => ({
+      copyArrGoods: state.copyArrGoods.reverse(),
```
Denchepornyuk, line 41: reverse is a mutable method, make a copy before using reverse

src/App.tsx
```diff
@@ -14,11 +15,136 @@ const goodsFromServer: string[] = [
...
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+const numArr: number[] = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+
+type State = {
+  visibleGoods: boolean,
+  copyArrGoods: string[],
+  secondCopyArr: string[],
```
Denchepornyuk, line 23: do you need copyArrGoods and secondCopyArr?
I believe you can use only one array for goods in state

Denchepornyuk (changes requested):

PR #906, opened 2022-06-05
list of goods

kitaminary:
[DEMO LINK](https://kitaminary.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
+import classNames from 'classnames';
 import React from 'react';
-import './App.css';
+import './App.scss';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+                      key={good}
+                      className="good__item subtitle is-3"
+                    >
+                      {good}
+                    </li>
+                  ))}
+                </ul>
+                <button
+                  type="button"
+                  onClick={() => this.setState(state => ({
```
iwanttosmile4u, line 65: It'll be better if you write seperate method - render(), for example, - and set this method in onClick.

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+                    goods: [...state.goods].reverse(),
+                  }))}
+                  className="button is-warning is-medium"
+                >
+                  Reverse
+                </button>
+                <button
+                  type="button"
+                  onClick={() => this.setState(state => ({
```
iwanttosmile4u, line 74: In this case also will be better create separate method befor render().

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+                      good1.localeCompare(good2)
+                    )),
+                  }))}
+                  className="button is-warning is-medium"
+                >
+                  Sort alphabetically
+                </button>
+                <button
+                  type="button"
+                  onClick={() => this.setState((state) => ({
```
iwanttosmile4u, line 85: The same problem.

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+                    goods: state.resetGoods,
+                  }))}
+                  className="button is-warning is-medium"
+                >
+                  Reset
+                </button>
+                <button
+                  type="button"
+                  onClick={() => this.setState(state => ({
```
iwanttosmile4u, line 94: And here the same problem

iwanttosmile4u (commented):
Good job! I just have one common recommendation)

PR #908, opened 2022-06-05
solution

lytvyshko:
- [DEMO LINK](https://lytvyshko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import { GoodsList } from './components/GoodsList';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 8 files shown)


iwanttosmile4u (commented):
Good job!) But when goods are reversed and i want to sort them, they are sorted in the opposite direction. And if i don't use reverse and just sort goods all works good. Is it ok?)

src/App.tsx
```diff
@@ -14,11 +15,126 @@ const goodsFromServer: string[] = [
...
+  reset = () => {
+    this.setState({
+      isReversed: false,
+      sortBy: '',
+    });
+  };
+
+  render() {
+    const { isVisible, isReversed, sortBy } = this.state;
+    const goodsCopy = goodsFromServer.slice();
```
Denchepornyuk, line 64: use spread operator to copy the array

denys-cheporniuk (approved):

PR #912, opened 2022-06-06
Develop

v-vytrykush:
[DEMO](https://v-vytrykush.github.io/react_list-of-goods/)

### DIFF
--- a/src/components/GoodsList.tsx
+++ b/src/components/GoodsList.tsx
@@ -27,7 +27,9 @@ export class GoodsList extends React.Component<Props, State> {
     return (
       this.setState(state => (
         {
-          goods: [...state.goods].sort(),
+          goods: [...state.goods]
+            .sort((firstGood, secondGood) => firstGood
+              .localeCompare(secondGood)),
         }
       ))
     );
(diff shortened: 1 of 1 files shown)


src/components/GoodsList.tsx
```diff
@@ -0,0 +1,103 @@
...
+        }
+      ))
+    );
+  };
+
+  sortGoodsAlphabetically = () => {
+    return (
+      this.setState(state => (
+        {
+          goods: [...state.goods].sort(),
```
anastasiia-nikita, line 30: Better to use locale.Compare()

anastasiia-nikita (approved):
Work good!👍

src/components/GoodsList.tsx
Denchepornyuk: agree, use localeCompare for sorting strings

src/App.tsx
```diff
@@ -14,11 +15,34 @@ const goodsFromServer: string[] = [
...
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    startBtn: false,
+  };
+
+  start = () => {
+    return (this.setState({ startBtn: true })
+    );
```
Denchepornyuk, line 29:
```suggestion
    return this.setState({ startBtn: true });
```

Denchepornyuk (approved):

PR #919, opened 2022-06-09
My solve

Yaroslav-Kononenko:
[DEMO LINK](https://Yaroslav-Kononenko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import { GoodsList } from './components/GoodsList';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -14,11 +15,127 @@ const goodsFromServer: string[] = [
...
+  isReversed: boolean,
+  sortBy: string,
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    goods: goodsFromServer,
+    isVisible: false,
+    isReversed: false,
+    sortBy: '',
```
Denchepornyuk, line 30: you don't have to save these values in state, make reverse and when on button click

src/App.tsx
```diff
@@ -14,11 +15,127 @@ const goodsFromServer: string[] = [
...
+  sortByAlfabeth = () => {
+    this.setState({
+      sortBy: 'alphabet',
+    });
+  };
+
+  reverse = () => {
+    this.setState(state => ({
+      isReversed: !state.isReversed,
+    }));
```
Denchepornyuk, line 50:
```suggestion
    this.setState(state => ({
      goods: [...state.goods].reverse(),
    }));
```
the same for sort

src/App.tsx
```diff
@@ -14,11 +15,127 @@ const goodsFromServer: string[] = [
...
+      sortBy,
+    } = this.state;
+    const visibleGoods = [...goods];
+
+    if (sortBy === 'length') {
+      visibleGoods.sort((good1, good2) => good1.length - good2.length);
+    }
+
+    if (sortBy === 'alphabet') {
+      visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
```
Denchepornyuk, line 80: sort and reverse are mutable methods so you need to make copy before using these methods

Denchepornyuk (changes requested):

Anna-Tkachenko (approved):
good job! 👍

PR #920, opened 2022-06-09
simple tasks solution

efdalyalcin:
- [DEMO LINK](https://efdalyalcin.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -1,1 +1,14 @@
-/* not empty */
+@import "https://cdn.jsdelivr.net/npm/bulma@0.9.4/css/bulma.min.css";
+
+.App__start {
+  height: 48px;
+  margin: 16px;
+}
+
+.button {
+  margin: 8px;
+}
+
+.container {
+  margin: 0 auto;
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,11 +14,140 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+interface State {
+  goods: string[];
+  isStarted: boolean;
+  isReversed: boolean;
+  sortBy: 'none' | 'name' | 'length';
+}
+
+class App extends React.Component<{}, State> {
+  state: State = {
+    goods: [...goodsFromServer],
+    isStarted: false,
+    isReversed: false,
+    sortBy: 'none',
+  };
+
+  start = () => {
+    this.setState({ isStarted: true });
+  };
+
+  reverse = () => {
+    this.setState((state) => (
+      { isReversed: !state.isReversed }
+    ));
+  };
+
+  sortByAlpha = () => {
+    this.setState({ sortBy: 'name' });
+  };
+
+  sortByLength = () => {
+    this.setState({ sortBy: 'length' });
+  };
+
+  reset = () => {
+    this.setState({ isReversed: false, sortBy: 'none' });
+  };
+
+  render(): React.ReactNode {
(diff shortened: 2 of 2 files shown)


LukaszGryziewicz (approved):
Good job! 👍

src/App.tsx
```diff
@@ -14,11 +14,196 @@ const goodsFromServer: string[] = [
...
-    {goodsFromServer.length}
-  </div>
-);
+const lengths: number[] = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+
+interface State {
+  goods: string[];
+  isStarted: boolean;
+  isReversed: boolean;
+  sortBy: 'none' | 'name' | 'length';
```
meforoma, line 23: let's make it enum and use in code instead of raw strings
```
enum SortBy {
  none,
  name,
  length,
}
```

```suggestion
  sortBy: SortBy;
```

src/App.tsx
```diff
@@ -14,11 +14,196 @@ const goodsFromServer: string[] = [
...
+  selection: number;
+}
+
+class App extends React.Component<{}, State> {
+  state: State = {
+    goods: [...goodsFromServer],
+    isStarted: false,
+    isReversed: false,
+    sortBy: 'none',
```
meforoma, line 32:
```suggestion
    sortBy: SortBy.none,
```
and for all strings along the code

src/App.tsx
```diff
@@ -14,11 +14,196 @@ const goodsFromServer: string[] = [
...
+  };
+
+  hide = () => (this.setState(state => (
+    {
+      isStarted: !state.isStarted,
+      isReversed: false,
+      sortBy: 'none',
+      selection: 1,
+    }
+  )));
```
meforoma, line 47: i might be wrong,
but it looks like you're not only hiding,
but rather `toggleVisibility`
`isStarted: !state.isStarted` - this swaps the state
```suggestion
  toggleVisibility = () => (this.setState(state => (
    {
      isStarted: !state.isStarted,
      isReversed: false,
      sortBy: 'none',
      selection: 1,
    }
  )));
```

meforoma: also you seem to also reset all the setting.
Could be generally, but you already have reset as a separate method.
So let's not repeat the functionality
```suggestion
  toggleVisibility = () => (this.setState(state => (
    {
      isStarted: !state.isStarted,
    }
  )));
```

meforoma (changes requested):
cool implementation @efdalyalcin 
let's make it even better.

src/App.css
```diff
@@ -1 +1,23 @@
...
+  margin: 8px;
+}
+
+.container {
+  margin: 0 auto;
+}
+
+.centered {
+  margin: 0 auto;
+}
```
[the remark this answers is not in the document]
efdalyalcin: Yes I also figured it out by myself, after some point it became impossible to make it right 😃 I totally agree with you but for this one I don't want to loose more time on styling 🥹

src/App.tsx
efdalyalcin: all done...

src/App.tsx
```diff
@@ -14,11 +14,202 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+const lengths: number[] = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
```
ihor-jpeg, line 17: 🤯 
```suggestion
const lengths: number[] = [...Array(11).keys()].slice(1);
```

ihor-jpeg (approved):

PR #921, opened 2022-06-09
React List of Goods Solution

59MrRobot:
- [DEMO LINK](https://59MrRobot.github.io/react_list-of-goods/)
(havent deployed yet)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,11 +14,140 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type SortBy = 'none' | 'name' | 'length';
+
+type State = {
+  goods: string[];
+  isVisible: boolean;
+  isReversed: boolean,
+  sortBy: SortBy,
+};
+
+class App extends React.Component<{}, State> {
+  state: State = {
+    goods: [...goodsFromServer],
+    isVisible: false,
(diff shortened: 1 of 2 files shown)


efdalyalcin (approved):
🔥🔥🔥

src/App.tsx
```diff
@@ -14,11 +14,140 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Goods</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type SortBy = 'none' | 'name' | 'length';
```
meforoma, line 17: let's make it an enum and use in code
like
```suggestion
enum SortBy {
  none,
  name,
  length,
}
```

and use `SortBy.none` etc.

meforoma (changes requested):

src/App.tsx
59MrRobot: Alright, will do 👍

src/App.tsx
```diff
@@ -14,11 +14,154 @@ const goodsFromServer: string[] = [
...
+    ));
+  };
+
+  sortByName = () => {
+    this.setState(state => ({
+      goods: [...state.goods].sort((good1, good2) => (
+        (good1).localeCompare(good2)
+      )),
+    }));
+  };
```
ihor-jpeg, line 48: maybe store sorting criteria in the state and avoid doing the same thing (sorting) twice?

ihor-jpeg (changes requested):

PR #931, opened 2022-06-13
task started and added reverse

Sardorbek1122:
[DEMO LINK](https://Sardorbek1122.github.io/react_list-of-goods/)


LukaszGryziewicz (approved):
Nice work, next time I would recommend creating an enum for the types of sorting

Sardorbek1122:
> Nice work, next time I would recommend creating an enum for the types of sorting

thank you

meforoma (approved):

PR #932, opened 2022-06-13
process

Uktam19980416:


efdalyalcin (changes requested):
I think you couldn't push and deploy

src/App.tsx
```diff
@@ -14,11 +14,193 @@ const goodsFromServer: string[] = [
...
+    this.setState({
+      goods: [...goodsFromServer],
+      isReversed: false,
+      isSorted: Sorting.Default,
+      length: 1,
+    });
+  };
+
+  sortByReverse = () => {
+    this.setState({ isReversed: true });
```
meforoma, line 54: now your `reverse` is a one-time use
enable it to reverse every time it's clicked - use 
```suggestion
    this.setState({ isReversed: !this.state.isReversed });
```

meforoma (changes requested):
found the deployment, but would be nice of you to provide the link
https://uktam19980416.github.io/react_list-of-goods/

Uktam19980416:
[DEMO LINK](https://Uktam19980416.github.io/react_list-of-goods/)

src/App.tsx
```diff
@@ -14,11 +14,195 @@ const goodsFromServer: string[] = [
...
+                  <option value="1">1</option>
+                  <option value="2">2</option>
+                  <option value="3">3</option>
+                  <option value="4">4</option>
+                  <option value="5">5</option>
+                  <option value="6">6</option>
+                  <option value="7">7</option>
+                  <option value="8">8</option>
+                  <option value="9">9</option>
+                  <option value="10">10</option>
```
efdalyalcin, line 183: you could have created an array and use forEach so that you won't copy paste each option...

PR #949, opened 2022-07-11
Solution

stanislav-topikha:
[DEMO LINK](https://stanislav-topikha.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import { GoodsList } from './components/GoodsList';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


vsupruniuk (approved):

src/App.tsx
```diff
@@ -14,11 +15,42 @@ const goodsFromServer: string[] = [
...
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  listShown: boolean,
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    listShown: false,
```
Olexandr6, line 24:
```suggestion
    isListShown: false,
```

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,149 @@
...
+          return g1.length - g2.length;
+
+        default:
+          return 0;
+      }
+    });
+
+    if (isReverse) {
+      preparedList.reverse();
+    }
```
Olexandr6, line 68: This logic shouldn't be here. Method `render` should be as clear as possible.
Let's move this logic to a separate method

Olexandr6 (changes requested):

PR #950, opened 2022-07-11
solution

kostya-konev:
[DEMO LINK](https://kostya-konev.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import './App.css';
+import React, { useState } from 'react';
+import './App.scss';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -14,11 +14,132 @@ const goodsFromServer: string[] = [
...
+            <button
+              type="button"
+              className="button"
+              onClick={() => setIsShown(true)}
+            >
+              Start
+            </button>
+          ) : (
+            <div>
+              <ul className="App__list">
```
alexandra-protyanova, line 55: Create separate component for GoodsList

src/App.tsx
```diff
@@ -14,11 +14,132 @@ const goodsFromServer: string[] = [
...
+              </div>
+              <div className="App__select">
+                <span className="App__select--title">
+                  Choose length
+                </span>
+                <select
+                  value={length}
+                  onChange={event => setLength(event.target.value)}
+                >
+                  <option value="1">
```
alexandra-protyanova, line 106: u can create array from 1 to 10 and then map it to create options

src/App.tsx
```diff
@@ -14,11 +14,132 @@ const goodsFromServer: string[] = [
...
+      .filter(good => good.length >= +length));
+  };
+
+  const reset = () => {
+    setGoods(goodsFromServer);
+    setLength('1');
+  };
+
+  const sortByLength = () => {
+    setGoods([...goods].sort((a, b) => b.length - a.length));
```
alexandra-protyanova, line 38: you have 2 functions with almost similar logic of sort, combine them

alexandra-protyanova (changes requested):

src/App.tsx
kostya-konev: What do you mean?

alexandra-protyanova: u can create a function that gonna accepts the type of sort and sort it in that case, store this type in state, and change it on click. In function it is better to use switch

alexandra-protyanova (approved):
It is not necessarily to create this thing, but it would be good)

src/App.tsx
kostya-konev: I don't want to do it

PR #955, opened 2022-07-12
Solution

liliia-livandovskaya:
[DEMO LINK](https://liliia-livandovskaya.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import './App.css';
+import './App.scss';
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -14,11 +14,114 @@ const goodsFromServer: string[] = [
...
+  visabilityToggle = () => {
+    this.setState(state => ({
+      isVisible: !state.isVisible,
+    }));
+  };
+
+  reverse = () => {
+    this.setState(state => ({
+      goodsList: [...state.goodsList].reverse(),
+    }));
```
AndriiNychaiuk, line 37: Great job! Your solution is quite good! But there is one thing I'm not sure. Every your action is creating new array. But you alredy created(copied) it in your initial state. Isn't it way better to work with it instead and create the new one only if it's really necessary?

AndriiNychaiuk (changes requested):
Your list is too big. It is very inconvenient to scroll through it to see the whole result after pressing the buttons

AndriiNychaiuk (approved):
👍

src/App.tsx
```diff
@@ -14,11 +14,114 @@ const goodsFromServer: string[] = [
...
+  reset = () => {
+    this.setState({ goodsList: [...goodsFromServer] });
+  };
+
+  sortByLength = () => {
+    this.setState(state => ({
+      goodsList: state.goodsList.sort((a, b) => a.length - b.length),
+    }));
+  };
+
```
mykola-koval, line 55: can combine this into a single function with switch/case

src/App.tsx
```diff
@@ -14,11 +14,114 @@ const goodsFromServer: string[] = [
...
+                  {goodsList.map(
+                    good => (
+                      <li key={good}>
+                        {good}
+                      </li>
+                    ),
+                  )}
+                  {' '}
+                </ul>
+                <button
```
mykola-koval, line 76: add one empty line between same level elements
```suggestion
                </ul>
                
                <button
```

mykola-koval (approved):

PR #989, opened 2022-07-15
Solution

Dmytrit:
[DEMO LINK](https://Dmytrit.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,55 +21,119 @@ enum SortType {
   LENGTH,
 }
 
-// Use this function in the render method
-function getReorderedGoods(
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -21,55 +21,119 @@ enum SortType {
...
+                this.setState({ isStarted: true });
+              }}
+            >
+              Start
+            </button>
+          </div>
+
+        )}
+
+        {isStarted && (
```
jstmpelowycz, line 78: I'd rather use ternary here

jstmpelowycz (approved):

PR #990, opened 2022-07-16
Solution

AntonYegorov:
[DEMO LINK](https://AntonYegorov.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,7 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
 import React from 'react';
 import './App.css';
 
-const goodsFromServer = [
+const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


vsupruniuk (approved):
👍🔥

illiarerun (changes requested):

src/App.tsx
```diff
@@ -21,55 +20,140 @@ enum SortType {
...
+            return good1.localeCompare(good2);
+
+          default: return 0;
+        }
+      });
+    }
+
+    if (isReversed) {
+      newGoods.reverse();
+    }
```
jstmpelowycz, line 94: this logic is for a separate function
just to explain you why: combining logic of several separate processes in one method (in this case – `render`) is not a common practice, it makes your unmaintainable

jstmpelowycz: strongly recommend you using this approach next times

jstmpelowycz (approved):

PR #1001, opened 2022-07-18
finished list of goods

omelika:
[DEMO LINK](https://omelika.github.io/react_list-of-goods/)


src/App.tsx
```diff
@@ -17,59 +17,112 @@ const goodsFromServer = [
 
 enum SortType {
   NONE,
-  ALPABET,
+  ALPHABET,
   LENGTH,
 }
```
yanaost, line 23: you deleted function getReorderGoods...
As far as I get it we need to use it for main functionality

yanaost (approved):
you deleted function getReorderGoods...
As far as I get it we need to use it for main functionality:
we call the function to reorder items with state elements as arguments;
when state changes,  function will be called one more time

src/App.tsx
illiarerun: Agree with Yana here

src/App.tsx
```diff
@@ -17,59 +17,112 @@ const goodsFromServer = [
...
+          return good1.length - good2.length;
+
+        default:
+          return 0;
+      }
+    });
+
+    if (isReversed) {
+      goods.reverse();
+    }
```
illiarerun, line 80: move this logic to the function that Yana has mentioned and just call it here

illiarerun (changes requested):

mykola-koval (approved):

PR #1005, opened 2022-07-18
Solution

dimabelotskyi:
[DEMO LINK](https://dimabelotskyi.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import React from 'react';
+import { FC, useState } from 'react';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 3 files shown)


ant2nta (approved):
gj

ant2nta (commented):
you should have done it with class component

src/App.tsx
```diff
@@ -16,60 +16,97 @@ const goodsFromServer = [
...
+  visibleGoods.sort((firstGood, secondGood) => {
+    switch (sortBy) {
+      case SortType.Alpabet:
+        return firstGood.localeCompare(secondGood);
+      case SortType.Length:
+        return firstGood.length - secondGood.length;
+      default:
+        return 0;
+    }
+  });
```
jstmpelowycz, line 41: I would suggest one more option you could have used: just create handlers and pass them to buttons, and when those button are clicked – perform the action (i.e. sort, reset, show – and so on)

why: every time your component gets re-rendered – you perform this action, is this okay?

jstmpelowycz (approved):

PR #1014, opened 2022-07-27
solution

FANTAZER-NURE:
https://FANTAZER-NURE.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,55 +21,140 @@ enum SortType {
   LENGTH,
 }
 
-// Use this function in the render method
-function getReorderedGoods(
-  goods: string[],
-  sortType: SortType,
-  isReversed: boolean,
-) {
-  // Not to mutate the original array
-  const visibleGoods = [...goods];
-
-  // Sort and reverse goods if needed
-  // ...
-
-  return visibleGoods;
-}
-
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -21,55 +21,140 @@ enum SortType {
...
+    if (sortType === SortType.LENGTH) {
+      goods = goods.sort((a, b) => a.length - b.length);
+    }
+
+    if (sortType === SortType.NONE) {
+      goods = [...goodsFromServer];
+    }
+
+    if (isReversed) {
+      goods.reverse();
```
alexandra-protyanova, line 84: use switch case here

src/App.tsx
```diff
@@ -21,55 +21,140 @@ enum SortType {
...
+              <button
+                type="button"
+                onClick={this.start}
+                className="button is-light is-primary level-item"
+              >
+                Start
+              </button>
+            )}
+
+          {this.state.isStarted
```
alexandra-protyanova, line 101: it is better to create separate component for goods and appropriate buttons

alexandra-protyanova (changes requested):

valerii-derkach (approved):

PR #1019, opened 2022-08-03
Solution

fledith:
[DEMO LINK](https://fledith.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
-import React from 'react';
+import { Component } from 'react';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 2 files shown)


BudnikOleksii (approved):
Great!

src/App.tsx
```diff
@@ -15,61 +14,185 @@ const goodsFromServer = [
...
-      <li className="Goods__item">...</li>
-    </ul>
-  </div>
-);
+export class App extends Component<{}, State> {
+  state = {
+    isStarted: false,
+    isReversed: false,
+    sortType: SortType.NONE,
+    selected: 1,
```
solaryasha, line 39: Let's change it to max value, so user could see something

src/App.tsx
```diff
@@ -15,61 +14,185 @@ const goodsFromServer = [
...
+  );
+
+  sortAlphabetically = () => (
+    this.setState((state) => ({
+      sortType: SortType.ALPHABET,
+      isSortedAlphabetically: !state.isSortedAlphabetically,
+    }))
+  );
+
+  revers = () => (
```
solaryasha, line 93:
```suggestion
  reverse = () => (
```

src/App.tsx
```diff
@@ -15,61 +14,185 @@ const goodsFromServer = [
...
+    this.setState(state => ({ isReversed: !state.isReversed }))
+  );
+
+  handleChangeSelect = (event: ChangeEvent<HTMLSelectElement>) => (
+    this.setState({ selected: +event.target.value })
```
solaryasha, line 98:
```suggestion
    this.setState({ selected: Number(event.target.value) })
```

src/App.tsx
```diff
@@ -15,61 +14,185 @@ const goodsFromServer = [
...
+                  <option value={length} key={length}>
+                    {length}
+                  </option>
+                ))}
+              </select>
+            </div>
+
+            <ul className="Goods">
+              {goods.map(good => (
+                <li className="Goods__item" key={good}>
```
solaryasha, line 188: What if we would have duplicated goods in the array ?

solaryasha (changes requested):

src/App.tsx
fledith: I agree if you take into account the empty line, but in another way, I think that better used the unary operator

src/App.tsx
fledith: I know, I was just too lazy to download the library key generator

solaryasha (approved):

PR #1032, opened 2022-08-09
ready

denenergi:
[DEMO LINK](https://denenergi.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import React from 'react';
+import { PureComponent } from 'react';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -27,11 +27,24 @@ function getReorderedGoods(
...
+      switch (sortType) {
+        case SortType.ALPABET:
+          return a.localeCompare(b);
+        case SortType.LENGTH:
+          return a.length - b.length;
+        default:
+          return 0;
+      }
+    });
+  }
```
ivan-kadykalo, line 43: It is good when you can avoid nested if statements

```suggestion
    switch (sortType) {
      case SortType.ALPABET:
        return visibleGoods.sort((a, b) => a.localeCompare(b));
      case SortType.LENGTH:
        return visibleGoods.sort((a, b) => a.length - b.length);
      case SortType.NONE:
      default:
        break;
    }
```

ivan-kadykalo (approved):

PR #1041, opened 2022-08-10
Solution

LeonidPustovit:
[DEMO LINK](https://LeonidPustovit.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
 import React from 'react';
+import classNames from 'classnames';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -41,35 +59,134 @@ type State = {
...
+                </select>
+              </div>
+
+              <ul className="Goods">
+                {getReorderedGoods(
+                  goodsFromServer,
+                  sortType,
+                  isReversed,
+                  minLength,
+                ).map(good => {
```
Ihor-Karpyn, line 171: move to var

Ihor-Karpyn (approved):
Good job, don't forget to mark the active button to explain to the user what action is active

Ihor-Karpyn (changes requested):
.

src/App.tsx
```diff
@@ -26,12 +26,30 @@ function getReorderedGoods(
   goods: string[],
   sortType: SortType,
   isReversed: boolean,
+  minLength: number,
 ) {
   // Not to mutate the original array
-  const visibleGoods = [...goods];
+  const visibleGoods = [...goods].filter(good => good.length > minLength);
```
solaryasha, line 32: Method filter doesn't mutate the array, you don't need to create copy here

solaryasha (approved):

PR #1042, opened 2022-08-10
Solution

rodiontseva:
[DEMO LINK](https://rodiontseva.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,55 +21,121 @@ enum SortType {
   LENGTH,
 }
 
-// Use this function in the render method
-function getReorderedGoods(
-  goods: string[],
-  sortType: SortType,
-  isReversed: boolean,
-) {
-  // Not to mutate the original array
-  const visibleGoods = [...goods];
-
-  // Sort and reverse goods if needed
-  // ...
-
-  return visibleGoods;
-}
-
-// DON'T save goods to the state
 type State = {
   isStarted: boolean,
   isReversed: boolean,
   sortType: SortType,
 };
 
-export const App = () => (
-  <div className="App">
-    <button type="button">
-      Start
-    </button>
+function getReorderedGoods(
+  goods: string[],
+  sortType: SortType,
+  isReversed: boolean,
+) {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -21,55 +26,143 @@ enum SortType {
...
-  // ...
+  visibleGoods.sort((good1, good2) => {
+    switch (sortType) {
+      case SortType.ALPABET:
+        return good1.localeCompare(good2);
+
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      case SortType.NONE:
```
roman-romanenko, line 50: You don't need this case 
As for me

roman-romanenko (approved):
Good job 👍🏻

src/App.tsx
```diff
@@ -21,55 +26,143 @@ enum SortType {
...
+      default:
+        return 0;
+    }
+  });
```
ivan-kadykalo, line 54: Your sort will start every time and will check each element and do nothing
```suggestion
    switch (sortType) {
      case SortType.ALPABET:
        return visibleGoods.sort((a, b) => a.localeCompare(b));
      case SortType.LENGTH:
        return visibleGoods.sort((a, b) => a.length - b.length);
      case SortType.NONE:
      default:
        break;
    }
```

ivan-kadykalo (changes requested):

src/App.tsx
rodiontseva: working this way:
  switch (sortType) {
    case SortType.ALPABET:
      visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
      **break;**
    case SortType.LENGTH:
      visibleGoods.sort((good1, good2) => good1.length - good2.length);
      **break;**
    case SortType.NONE:
    default:
      break;
  }

rodiontseva:
> Good job, don't forget to mark the active button to explain to the user what action is active

done https://rodiontseva.github.io/react_list-of-goods/

PR #1045, opened 2022-08-10
list sort reverse solution

roman-romanenko:
[DEMO LINK](https://roman-romanenko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import React from 'react';
+import { Component } from 'react';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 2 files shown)


rodiontseva (approved):
great! and nice style!

src/App.tsx
```diff
@@ -21,17 +21,27 @@ enum SortType {
...
+  visibleGoods.sort((g1, g2) => {
+    switch (sortType) {
+      case SortType.ALPABET:
+        return g1.localeCompare(g2);
+      case SortType.LENGTH:
+        return g1.length - g2.length;
+      default:
+        return 0;
+    }
+  });
```
ivan-kadykalo, line 40: Your sort will start every time and will check each element and do nothing
```suggestion
  switch (sortType) {
      case SortType.ALPABET:
        return visibleGoods.sort((a, b) => a.localeCompare(b));
      case SortType.LENGTH:
        return visibleGoods.sort((a, b) => a.length - b.length);
      case SortType.NONE:
      default:
        break;
    }
```

ivan-kadykalo (changes requested):

src/App.tsx
roman-romanenko: thanks , good point. Additionaly I delete return from switch because , revers didn't work after sort

Ihor-Karpyn (approved):
Good job, don't forget to mark the active button to explain to the user what action is active

ivan-kadykalo (approved):

PR #1046, opened 2022-08-10
solution

Vova-Balyshev:
[DEMO LINK](https://Vova-Balyshev.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,8 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
+
 import React from 'react';
+import Button from '@mui/material/Button';
+import Stack from '@mui/material/Stack';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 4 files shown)


Yura-Lysenko (approved):
looks nice, well done!

src/App.tsx
```diff
@@ -21,55 +24,135 @@ enum SortType {
...
+        case SortType.ALPABET:
+          return good1.localeCompare(good2);
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+
+        default:
+          return 0;
+      }
+    });
+  }
```
ivan-kadykalo, line 46: It is good when you can avoid nested if, it will be easier to read
```suggestion

  switch (sortType) {
    case SortType.ALPABET:
      return visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
    case SortType.LENGTH:
      return visibleGoods.sort((good1, good2) => good1.length - good2.length);
    case SortType.NONE:
    default:
      break;
  }
```

ivan-kadykalo (approved):

PR #1047, opened 2022-08-10
Solution

kata96451:
[DEMO LINK](https://kata96451.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import React from 'react';
+import { Component } from 'react';
 import './App.css';
 
 const goodsFromServer = [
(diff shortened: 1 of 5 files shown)


katerinahradinarova (approved):
well done)

src/App.tsx
```diff
@@ -17,59 +17,139 @@ const goodsFromServer = [
...
+        case SortType.ALPHABET:
+          return f1.localeCompare(f2);
+
+        case SortType.LENGTH:
+          return f1.length - f2.length;
+
+        default:
+          return 0;
+      }
+    });
```
ivan-kadykalo, line 43: It is good practice when you can avoid nested ifs. Code will be cleaner and will have better readability
```suggestion
      switch (sortType) {
        case SortType.ALPHABET:
          return visibleGoods.sort((f1, f2) => f1.localeCompare(f2));
        case SortType.LENGTH:
          return visibleGoods.sort((f1, f2) => f1.length - f2.length);
        case SortType.NONE:
        default:
          return 0;
      }
```

ivan-kadykalo (approved):
🚀

PR #1048, opened 2022-08-10
Implemented list of goods

illnino380:
[DEMO LINK](https://illnino380.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
 import React from 'react';
 import './App.css';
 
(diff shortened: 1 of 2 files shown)


katerinahradinarova (approved):
nice 🤩

src/App.tsx
```diff
@@ -15,61 +14,159 @@ const goodsFromServer = [
...
 
-  return visibleGoods;
+        case SortType.LENGTH:
+          return firstGood.length - secondGood.length;
+
+        default:
+          return 0;
+      }
+    });
+  }
```
ivan-kadykalo, line 58: It is good when you can avoid nested if statements


```
    switch (sortType) {
      case SortType.ALPABET:
      case SortType.LENGTH:
      case SortType.NONE:
      default:
        break;
    }

ivan-kadykalo (approved):

src/App.tsx
illnino380: I just thought not to start sorting if the state isn't changed. But I redo

PR #1051, opened 2022-08-11
Solution

alexbardyshev:
[DEMO LINK](https://alexbardyshev.github.io/react_list-of-goods/)


Ihor-Karpyn (changes requested):
Code is ok, but let's improve UI, highlight active button. If I click to sort by the length I want to know which sort is active now. The same with reverse

Ihor-Karpyn (approved):

PR #1059, opened 2022-08-11
solution

denis-khristyuk:
[DEMO LINK](https://denis-khristyuk.github.io/react_list-of-goods/)


Ihor-Karpyn (approved):
Good job, good practice is highlight active btn

PR #1073, opened 2022-08-29
solution

Kulchevych:
[DEMO LINK](https://Kulchevych.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import React from 'react';
 import './App.css';
+import { Good } from './components/Good/Good';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -27,11 +28,24 @@ function getReorderedGoods(
   sortType: SortType,
   isReversed: boolean,
 ) {
-  // Not to mutate the original array
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // ...
+  visibleGoods.sort((prev, next) => {
```
dellachaise, line 33: Variable names should be meaningful instead of `prev and next` you could use `product1 and product2` for example.

dellachaise (changes requested):
Great start 👍 

For now, sort works partly incorrect.
When I choose `sort alphabetically then `reverse`(everything great) but if then I choose `sort alphabetically` -  nothing changed(products didn't stay to right alphabetical order).

The same with `sort by length` it doesn't work correctly after reverse.

dellachaise (approved):
Great job 🔥

PR #1076, opened 2022-08-31
Task solution

sergii-nosachenko:
[DEMO LINK](https://sergii-nosachenko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,16 +37,18 @@ function getReorderedGoods(
 ) {
   const visibleGoods = [...goods].filter(good => good.length >= minLength);
 
-  visibleGoods.sort((a, b) => {
-    switch (sortType) {
-      case SortType.ALPABET:
-        return a.localeCompare(b);
-      case SortType.LENGTH:
-        return a.length - b.length;
-      default:
-        return 0;
-    }
-  });
+  if (sortType !== SortType.NONE) {
+    visibleGoods.sort((a, b) => {
+      switch (sortType) {
+        case SortType.ALPABET:
+          return a.localeCompare(b);
+        case SortType.LENGTH:
+          return a.length - b.length;
+        default:
+          return 0;
+      }
+    });
+  }
 
   if (isReversed) {
     visibleGoods.reverse();


src/App.tsx
```diff
@@ -21,55 +29,196 @@ enum SortType {
...
   goods: string[],
   sortType: SortType,
   isReversed: boolean,
+  minLength: number,
 ) {
-  // Not to mutate the original array
-  const visibleGoods = [...goods];
+  const visibleGoods = [...goods].filter(good => good.length >= minLength);
+
+  visibleGoods.sort((a, b) => {
```
mgrinko, line 40: if `sortType` is not set it is better to avoid sort call at all

src/App.tsx
```diff
@@ -21,55 +29,196 @@ enum SortType {
...
-      <li className="Goods__item">Dumplings</li>
-      <li className="Goods__item">Carrot</li>
-      <li className="Goods__item">Eggs</li>
-      <li className="Goods__item">...</li>
-    </ul>
-  </div>
-);
+const initialState: State = {
+  isStarted: false,
+  isInitial: true,
```
mgrinko, line 68: I think this value can be calculated from the others, so it duplicates other parts of state and may become out of sync at some point

I think you don't need to store it

src/components/Goods/Goods.tsx
```diff
@@ -0,0 +1,46 @@
...
+      >
+        {visibleGoods.map((good, index, goods) => (
+          <>
+            <ListItem
+              className="Goods__item"
+              key={`${good}`}
+            >
+              <ListItemText primary={good} />
+            </ListItem>
+            {index !== goods.length - 1 && <Divider />}
```
mgrinko, line 40: It is really hard to see that we have a tag here rendered conditionally
I would use () and put the tag on the next line

mgrinko (approved):
Please take my comments into account in the next tasks

src/App.tsx
sergii-nosachenko: @mgrinko This parameter `isInitial` controls look of Reset button. I've decided to create it because I need to know wether user has interacted with app. I can create one entry point for all user actions and modify it in one place, if it will be more complex application.

PR #1077, opened 2022-09-02
solution

vitaliy-kovbasinskyi:
[DEMO LINK](https://vitaliy-kovbasinskyi.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,11 +15,7 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-enum SortType {
-  NONE,
-  ALPABET,
-  LENGTH,
-}
+type SortType = 'none' | 'length' | 'name';
 
 // Use this function in the render method
 function getReorderedGoods(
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,11 +15,7 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-enum SortType {
```
alexandra-protyanova, line 18: I guess it's better use enum here, to be sure not to make a mistake is switch

src/App.tsx
```diff
@@ -43,33 +60,97 @@ type State = {
...
+              this.setState({
+                isStarted: true,
+              });
+            }}
+          >
+            Start
+          </button>
+        )}
+
+        {isStarted && (
```
alexandra-protyanova, line 94: I think it's better to create separate component for list of goods and it's buttons

alexandra-protyanova (approved):

PR #1085, opened 2022-09-09
solution

AlexanderKolomiiets:
[DEMO LINK](https://alexanderkolomiiets.github.io/react_list-of-goods/)


Dmytro-Matsehora (changes requested):
great work, Alexander, I offer only a small fix. Since you are already using deconstruction for `this.state`, I suggest using it for methods of `this` also

Dmytro-Matsehora (changes requested):
🔥 🔥 🔥

PR #1088, opened 2022-09-12
Develop

VolodymyrChuyko:
[DEMO LINK](https://VolodymyrChuyko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,10 @@
 import React from 'react';
+import classNames from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+import { ListOfGoods } from './components/ListOfGoods';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -35,6 +39,20 @@ export function getReorderedGoods(
   const visibleGoods = [...goods];
 
   // Sort and reverse goods if needed
+  switch (sortType) {
+    case SortType.ALPABET:
+      visibleGoods.sort((goodA, goodB) => goodA.localeCompare(goodB));
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((goodA, goodB) => goodA.length - goodB.length);
+      break;
+    default:
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -35,6 +39,20 @@ export function getReorderedGoods(
...
 
   // Sort and reverse goods if needed
+  switch (sortType) {
+    case SortType.ALPABET:
+      visibleGoods.sort((goodA, goodB) => goodA.localeCompare(goodB));
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((goodA, goodB) => goodA.length - goodB.length);
+      break;
+    default:
```
Jamal-Amoian, line 49: default value should return 0

src/App.tsx
```diff
@@ -35,6 +39,20 @@ export function getReorderedGoods(
...
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
```
Jamal-Amoian, line 57: Delete unnecessary console log

src/components/ListOfGoods/ListOfGoods.tsx
```diff
@@ -0,0 +1,19 @@
...
+};
+
+export class ListOfGoods extends React.PureComponent<Props> {
+  render() {
+    const { goods } = this.props;
+
+    return (
+      <ul>
+        {goods.map(good => (
+          <li data-cy="Good" key={Math.random()}>{good}</li>
```
Jamal-Amoian: You can use for key attribute good, 
`key={good}`

Jamal-Amoian (approved):
Well done!

src/App.tsx
VolodymyrChuyko: I like your solution to use a switch operator inside the callback. So I made changes to apply it to my code.

src/App.tsx
```diff
@@ -34,9 +38,17 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((goodA, goodB) => {
+    switch (sortType) {
+      case SortType.ALPABET: return goodA.localeCompare(goodB);
+      case SortType.LENGTH: return goodA.length - goodB.length;
```
denys-danyliuk, line 44:
```suggestion
      case SortType.ALPABET: 
        return goodA.localeCompare(goodB);
        
      case SortType.LENGTH: 
        return goodA.length - goodB.length;
```

it's hard to read, better to code this way

src/App.tsx
```diff
@@ -47,49 +59,78 @@ export function getReorderedGoods(
...
+              type="button"
+              className="button is-danger is-light"
+              onClick={this.handleReset}
+            >
+              Reset
+            </button>
+          )}
+        </div>
+
+        <ListOfGoods goods={getReorderedGoods(goodsFromServer, this.state)} />
```
denys-danyliuk, line 132: it will be better to save result of this function to variable and just pass it here than pass this function call

denys-danyliuk (changes requested):
Good job🚀
Just small issue, save result of function call to variable and pass it to other component

denys-danyliuk (approved):
GJ!

PR #1093, opened 2022-09-14
Develop

vladyslava-buzova:
[DEMO LINK](https://vladyslava-buzova.github.io/react_list-of-goods/)

Here is [the working version](https://mate-academy.github.io/react_list-of-goods/)

The button 'revers' works not the same as in the example. It was a bit complicated to discuss in the chat about it without seeing the code

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 import { GoodList } from './components/goods';
 
 export const goodsFromServer = [
@@ -27,6 +28,10 @@ type State = {
   sortType: boolean,
   sortBy: string,
   isShowm: boolean,
+  isClickedAlph: boolean,
+  isClickedLength: boolean,
+  isClickedReverse: boolean,
+  // toggleClass: boolean,
 };
 
 type ReorderOptions = {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -21,6 +23,17 @@ enum SortType {
...
 
+type State = {
+  isReversed: boolean,
+  sortType: boolean,
+  sortBy: string,
+  isShowm: boolean,
+  isClickedAlph: boolean,
+  isClickedLength: boolean,
+  isClickedReverse: boolean,
+  // toggleClass: boolean,
```
TarasHoliuk, line 34: please, don't leave comments

src/App.tsx
```diff
@@ -21,6 +23,17 @@ enum SortType {
...
 }
 
+type State = {
+  isReversed: boolean,
+  sortType: boolean,
+  sortBy: string,
+  isShowm: boolean,
+  isClickedAlph: boolean,
+  isClickedLength: boolean,
+  isClickedReverse: boolean,
```
TarasHoliuk, line 33: you can simplify it (decrease the count of properties)

src/App.tsx
```diff
@@ -34,62 +47,133 @@ export function getReorderedGoods(
...
+                'is-light': isClickedReverse,
+              })}
+            onClick={this.reverse}
+          >
+            Reverse
+          </button>
+
+          { isShowm
+          && (
+            <button
```
TarasHoliuk, line 165: This button shouldn't be shown if you clicked reverse 2 times (and no one sort is selected) - that causes the test to fail

TarasHoliuk (changes requested):
please simplify your app
hint: `sortBy` and `isReversed` in state will be enough

src/App.tsx
vladyslava-buzova: > This button shouldn't be shown if you clicked reverse 2 times (and no one sort is selected) - that causes the test to fail

there's the working version in the readme document, how it should work. 
Here is [the working version](https://mate-academy.github.io/react_list-of-goods/)

so I made it exactly as in this document.  Button reverse works absolutely the same way.
[DEMO LINK](https://vladyslava-buzova.github.io/react_list-of-goods/)
the rest of the buttons work the same way.

So I'm a bit confused about what instruction I should follow and how buttons should work

stanislav-topikha: Try to press reset 2 times on your version and on the working version
There is a difference

vladyslava-buzova:
https://vladyslava-buzova.github.io/react_list-of-goods/

src/App.tsx
```diff
@@ -15,81 +17,111 @@ export const goodsFromServer = [
...
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+  sortByLength = () => {
+    this.setState({
+      sortBy: 'length',
```
alexandra-protyanova, line 45: I guess enum was quite good idea here

alexandra-protyanova (approved):
Well done!

PR #1094, opened 2022-09-14
Solution

kir8mir:
- [DEMO LINK](https://kir8mir.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -34,6 +35,22 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((g1, g2) => (
+      g1.length - g2.length
+    ));
+  }
+
+  if (sortType === SortType.ALPABET || sortType === SortType.NONE) {
+    visibleGoods.sort((g1, g2) => (
+      g1.localeCompare(g2)
+    ));
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -34,6 +35,22 @@ export function getReorderedGoods(
...
+  if (sortType === SortType.ALPABET || sortType === SortType.NONE) {
+    visibleGoods.sort((g1, g2) => (
+      g1.localeCompare(g2)
+    ));
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
```
oksana-logos-frontend, line 53: better use switch here

src/App.tsx
```diff
@@ -34,6 +35,22 @@ export function getReorderedGoods(
...
   const visibleGoods = [...goods];
 
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((g1, g2) => (
+      g1.length - g2.length
+    ));
+  }
+
+  if (sortType === SortType.ALPABET || sortType === SortType.NONE) {
+    visibleGoods.sort((g1, g2) => (
```
oksana-logos-frontend, line 45: this solution make sort list in  alphabet by default it's wrong, because we have to do it on click, by default  our list have to be SortType.NONE

src/App.tsx
```diff
@@ -42,54 +59,101 @@ export function getReorderedGoods(
...
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+  resetBTN: boolean,
+};
+
+export class App extends Component<{}, State> {
+  state: State = {
+    isReversed: false,
+    sortType: SortType.ALPABET,
```
oksana-logos-frontend, line 71: we have to sort out list on click

oksana-logos-frontend (changes requested):
Good job, but add changes according to commits

oksana-logos-frontend (approved):
👍🏼

src/App.tsx
```diff
@@ -34,6 +35,26 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
+  switch (sortType) {
+    case SortType.LENGTH:
+      visibleGoods.sort((g1, g2) => (
```
AlinaMatuschak, line 40: Avoid abbreviations

src/App.tsx
```diff
@@ -34,6 +35,26 @@ export function getReorderedGoods(
...
+      ));
+      break;
+
+    case SortType.ALPABET:
+      visibleGoods.sort((g1, g2) => (
+        g1.localeCompare(g2)
+      ));
+      break;
+
+    default:
```
AlinaMatuschak, line 51: default can’t be empty. Should return default value or throw an error

src/App.tsx
```diff
@@ -42,54 +63,101 @@ export function getReorderedGoods(
...
+  state: State = {
+    isReversed: false,
+    sortType: SortType.NONE,
+    resetBTN: false,
+  };
+
+  render() {
+    const { isReversed, sortType, resetBTN } = this.state;
+    const myGoods = getReorderedGoods(goodsFromServer,
+      { sortType, isReversed });
```
AlinaMatuschak, line 82:
```suggestion
    const myGoods = getReorderedGoods(
      goodsFromServer,
      { sortType, isReversed },
    );
```

src/App.tsx
```diff
@@ -42,54 +63,101 @@ export function getReorderedGoods(
...
+  sortType: SortType,
+  resetBTN: boolean,
+};
+
+export class App extends Component<{}, State> {
+  state: State = {
+    isReversed: false,
+    sortType: SortType.NONE,
+    resetBTN: false,
+  };
```
AlinaMatuschak, line 77: Instead of using resetBTN, you can create a method that resets state to default values
```suggestion
  state: State = {
    isReversed: false,
    sortType: SortType.NONE,
  };
```

src/App.tsx
```diff
@@ -42,54 +63,101 @@ export function getReorderedGoods(
...
+            type="button"
+            className={classNames(
+              'button',
+              'is-info',
+              { 'is-light': sortType !== SortType.ALPABET },
+            )}
+            onClick={() => this.setState({
+              sortType: SortType.ALPABET,
+              resetBTN: true,
+            })}
```
AlinaMatuschak, line 97: Create separate methods instead of creating inline event handlers

AlinaMatuschak (changes requested):

danylo-onishchuk (approved):

PR #1148, opened 2022-09-21
solution

matvii1:
https://matvii1.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
+/* eslint-disable default-case */
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +29,121 @@ type ReorderOptions = {
...
+      isReversed: !state.isReversed,
+    }));
+  }
+
+  resetArr() {
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  }
```
liliia-livandovskaya, line 96: that's not required but it would be good to combine into a single function with switch/case

liliia-livandovskaya (approved):
Great job! 🔥

PR #1179, opened 2022-10-18
Implemented App component which render the list of goods and can sort or reversed it

Dikey945:
[DEMO LINK](https://Dikey945.github.io/react_list-of-goods/)


imaksweb (approved):
Nice work 👍

PR #1180, opened 2022-10-19
React list of goods

Vladosik1337:
[DEMO LINK](https://Vladosik1337.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -65,8 +65,30 @@ export class App extends React.Component<{}, State > {
     sortType: 0,
   };
 
+  sortByName = () => {
+    this.setState({ sortType: SortType.ALPABET });
+  };
+
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  reverseGoods = () => {
+    this.setState(state => ({
+      isReversed: !state.isReversed,
+    }));
+  };
+
+  reset = () => {
+    this.setState({
+      sortType: SortType.NONE,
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -38,58 +40,91 @@ export function getReorderedGoods(
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  } else if (sortType === SortType.ALPABET) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  }
+
```
nikitaovch99, line 48: If you have the same condition you better use the switch but in this case I think it's fine.

nikitaovch99 (approved):

src/App.tsx
```diff
@@ -38,58 +40,91 @@ export function getReorderedGoods(
...
+            className={classNames('button is-warning', {
+              'is-light': !isReversed,
+            })}
+            onClick={() => {
+              this.setState({ isReversed: !isReversed });
+            }}
+          >
+            Reverse
+          </button>
+          {(sortType !== SortType.NONE || isReversed) && (
```
silchyn, line 109: Better create variable to hold that condition

silchyn (changes requested):
Let's create callback functions for button handlers for readability

silchyn (approved):
Cool!

PR #1181, opened 2022-10-19
The ability to reorder the goods with the given buttons was implemented

Yevheniya-Deryka:
[DEMO LINK](https://Yevheniya-Deryka.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


Anna-Borodaienko (approved):

src/App.tsx
```diff
@@ -35,61 +36,123 @@ export function getReorderedGoods(
   const visibleGoods = [...goods];
 
   // Sort and reverse goods if needed
+  visibleGoods.sort((good1, good2) => {
+    switch (sortType) {
+      case 1:
```
Anna-Tkachenko, line 41: use enum here, for now it's not clear what these value means

Anna-Tkachenko (changes requested):

src/App.tsx
Yevheniya-Deryka: done!

Denchepornyuk (approved):

PR #1194, opened 2022-10-20
list of goods implemented

artem-musii:
[DEMO LINK](https://artem-musii.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { Component } from 'react';
+import classNames from 'classnames';
+import { v4 as uuid } from 'uuid';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -26,70 +28,109 @@ type ReorderOptions = {
...
+        return goodA.length - goodB.length;
+
+      default:
+        return 0;
+    }
+  });
+
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
```
TheVovchik, line 52: 👍

src/App.tsx
```diff
@@ -26,70 +28,109 @@ type ReorderOptions = {
...
+            )}
+            onClick={() => this.setState(state => ({
+              isReversed: !state.isReversed,
+            }))}
+          >
+            Reverse
+          </button>
+
+          {(isReversed
+            || sortType !== SortType.NONE) && (
```
TheVovchik, line 112: better to move this one in separate variable

src/App.tsx
```diff
@@ -26,70 +28,109 @@ type ReorderOptions = {
...
+            <button
+              type="button"
+              className="button is-danger is-light"
+              onClick={() => this.setState({
+                isReversed: false,
+                sortType: SortType.NONE,
```
TheVovchik, line 118: better to move this logic in method (and upper also)

TheVovchik (approved):
Leave some comments 
in line 112 better to create a separate value for condition

onClick actions better to move to a methods logic, so we can reuse them in the future (not a mistake offcourse)

in general great job 👍

artem-musii:
> Leave some comments in line 112 better to create a separate value for condition
> 
> onClick actions better to move to a methods logic, so we can reuse them in the future (not a mistake offcourse)
> 
> in general great job 👍

Thanks

src/App.tsx
```diff
@@ -26,70 +28,125 @@ type ReorderOptions = {
...
+              'button',
+              'is-warning',
+              { 'is-light': !isReversed },
+            )}
+            onClick={reverse}
+          >
+            Reverse
+          </button>
+
+          {(isChanged) && (
```
naziamoff, line 131: looks like those braces are unnecessary
```suggestion
          {isChanged && (
```

src/App.tsx
```diff
@@ -26,70 +28,125 @@ type ReorderOptions = {
...
+
+    const sortAlphabetically = () => {
+      this.setState({ sortType: SortType.ALPABET });
+    };
+
+    const sortByLength = () => {
+      this.setState({ sortType: SortType.LENGTH });
+    };
+
+    const reverse = () => {
```
naziamoff, line 79: should those be inside render method?

naziamoff (changes requested):

naziamoff (changes requested):

src/App.tsx
```diff
@@ -21,75 +23,130 @@ enum SortType {
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {reorderedGoods.map(good => (
+              <li key={uuid()} data-cy="Good">{good}</li>
```
naziamoff, line 145: key should be unique, but not random. With the key like that we will rerender the list over and over again, because those are always different list items

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #1199, opened 2022-10-20
Implement the ability to reorder the goods with the given buttons

OleksandrGlulhovskiy:
[DEMO LINK](https://OleksandrGlulhovskiy.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -38,58 +39,128 @@ export function getReorderedGoods(
...
-      <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {reorderedGoods.map(good => (
+            <li data-cy="Good" key={good}>
```
anastasiia-nahaiska, line 158: The mentor told me that we are not sure that the goods will never be repeated and therefore it is better to use [uuid](https://www.npmjs.com/package/uuid)
If you are going to do this, you need to prepare the array of goods in advance
image

anastasiia-nahaiska (approved):
GJ) Only one comment)

src/App.tsx
```diff
@@ -38,58 +39,128 @@ export function getReorderedGoods(
...
+      case SortType.LENGTH:
+        return first.length - second.length;
+
+      default:
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    return visibleGoods.reverse();
```
Anna-Tkachenko, line 56: return is redundant here

src/App.tsx
```diff
@@ -38,58 +39,128 @@ export function getReorderedGoods(
...
+          <button
+            type="button"
+            className={classNames(
+              'button is-success',
+              { 'is-light': sortType !== SortType.LENGTH },
+            )}
+            onClick={this.sortByLength}
+          >
+            Sort by length
+          </button>
```
Anna-Tkachenko, line 130: will be awesome to create separate component for buttons

Anna-Tkachenko (approved):

PR #1202, opened 2022-10-20
Solution

rstarovoit:
[DEMO LINK](https://rstarovoit.github.io/react_list-of-goods/) and add it to the PR description.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,70 +27,131 @@ type ReorderOptions = {
...
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((g1, g2) => {
```
grafit-off, line 36: Maybe, it's not critical, but what does `g` mean? I recommend use `prevGoods`, `currGoods` (or something like that) instead of `g`

src/App.tsx
```diff
@@ -26,70 +27,131 @@ type ReorderOptions = {
...
+              'button is-warning',
+              { 'is-light': !isReversed },
+            )}
+            onClick={this.reverseGoods}
+          >
+            Reverse
+          </button>
+
+          {
+            (sortType !== SortType.NONE || isReversed)
```
grafit-off, line 133: Probably, it's gonna be better to transform this condition into variable

grafit-off (approved):
Keep going ⭐

src/App.tsx
```diff
@@ -26,70 +27,129 @@ type ReorderOptions = {
...
+            ? (
+              <button
+                type="button"
+                className="button is-danger is-light"
+                onClick={this.resetSorting}
+              >
+                Reset
+              </button>
+            )
+            : null}
```
Andr1yK, line 142: in this case it is better to use &&

Andr1yK (approved):

PR #1203, opened 2022-10-20
list of goods was created

vallerii:
[DEMO LINK](https://vallerii.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -34,62 +35,116 @@ export function getReorderedGoods(
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {reorderedGoods.map(good => (
+              <li data-cy="Good">{good}</li>
```
korniienkokarina, line 143: you can add here key for map method `key={good}`

korniienkokarina (commented):
good job! left a comment below

src/App.tsx
```diff
@@ -34,62 +35,118 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((goodsA: string, goodsB: string) => {
```
naziamoff, line 38: You should give the variables proper names - they should describe what is stored inside. It shouldn't be in plural too in that case
```suggestion
  visibleGoods.sort((currentGood: string, nextGood: string) => {
```

src/App.tsx
```diff
@@ -34,62 +35,118 @@ export function getReorderedGoods(
...
+  alphabeticallySort = () => {
+    this.setState({ sortType: SortType.ALPABET });
+  };
+
+  lengtSort = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  reverse = () => {
+    this.setState(prevstate => ({
```
naziamoff, line 77: should be camelCased

naziamoff (approved):
🔥

PR #1204, opened 2022-10-21
sorted methods for list was added

iPanda177:
[DEMO LINK](https://iPanda177.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -17,7 +18,7 @@ export const goodsFromServer = [
 
 enum SortType {
   NONE,
-  ALPABET,
+  ALPHABET,
   LENGTH,
 }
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,130 @@ type ReorderOptions = {
...
+      case 2:
+        return item1.length - item2.length;
+
+      default:
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    return visibleGoods.reverse();
```
iam-datsenko, line 50: U don't need return here

src/App.tsx
```diff
@@ -26,70 +27,130 @@ type ReorderOptions = {
...
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((item1, item2) => {
+    switch (sortType) {
+      case 1:
+        return item1.localeCompare(item2);
+
+      case 2:
```
iam-datsenko, line 41: I think it's better to name cases according to the action they perform

src/App.tsx
```diff
@@ -26,70 +27,130 @@ type ReorderOptions = {
...
+    );
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={classNames(
+              'button is-info',
+              { 'is-light': sortType !== 1 },
```
iam-datsenko, line 102: Now it's hard to understand what is 1

iam-datsenko (approved):
GJ, but need some fixes

src/App.tsx
iPanda177: it won't work

src/App.tsx
iPanda177: good point, but i thought that will be more... accuracy)
but if mentors will say remake, i will) thx)

src/App.tsx
```diff
@@ -26,70 +27,130 @@ type ReorderOptions = {
...
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((item1, item2) => {
+    switch (sortType) {
+      case 1:
```
Anna-Tkachenko, line 38: use enum values here

src/App.tsx
Anna-Tkachenko: agree, don't need return here

src/App.tsx
```diff
@@ -26,70 +27,130 @@ type ReorderOptions = {
...
-      </div>
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    isReversed: false,
+    sortType: 0,
```
Anna-Tkachenko, line 64: set value from enum here

src/App.tsx
```diff
@@ -26,70 +27,130 @@ type ReorderOptions = {
...
+  };
+
+  sortAlphabet = () => {
+    this.setState({ sortType: 1 });
```
Anna-Tkachenko, line 68: same here, use enum

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -26,70 +27,130 @@ type ReorderOptions = {
...
+    sortType: SortType.NONE,
+  };
+
+  sortAlphabet = () => {
+    this.setState({ sortType: SortType.ALPHABET });
+  };
+
+  sortLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
```
Denchepornyuk, line 73: you can create one function for sortType, pass in arguments the value of SortType (ALPHABET, LENGTH) and set this value to the sortType

Denchepornyuk (approved):

PR #1205, opened 2022-10-21
sort goods implemented v.1

korniienkokarina:
[DEMO LINK](https://korniienkokarina.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -17,7 +18,7 @@ export const goodsFromServer = [
 
 enum SortType {
   NONE,
-  ALPABET,
+  ALPHABET,
   LENGTH,
 }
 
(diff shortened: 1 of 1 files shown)


vallerii:
GJ 👍

src/App.tsx
```diff
@@ -26,70 +27,123 @@ type ReorderOptions = {
...
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    });
+  };
+
+  render() {
+    const goods = getReorderedGoods(goodsFromServer, this.state);
+
+    const notChanged = this.state.sortType === SortType.NONE
```
naziamoff, line 90: Boolean variables should start with `is...`, `should...` or sth like that. We should follow those conventions in order to simplify the reading of our code
```suggestion
    const isNotChanged = this.state.sortType === SortType.NONE
```

src/App.tsx
```diff
@@ -26,70 +27,123 @@ type ReorderOptions = {
...
+            type="button"
+            onClick={this.handleReverseSort}
+            className={classNames('button is-warning', {
+              'is-light': !this.state.isReversed,
+            })}
+          >
+            Reverse
+          </button>
+
+          {!notChanged && (
```
naziamoff, line 126: actually it is quite difficult to process in my head, I read it as `not not changed`. Better to use positive condition here. Let's make the variable like `areGoodsChanged` or `areGoodsOriginal`

src/App.tsx
```diff
@@ -26,70 +27,123 @@ type ReorderOptions = {
...
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((goodA, goodB) => {
```
naziamoff, line 36: not the best idea to use A or B in the variable name. Better to come up with `currentGood`/`nextGood`. 
Don't think it is critical here though

naziamoff (approved):
Well done, there are just some naming mistakes which I recommend you to fix

PR #1206, opened 2022-10-21
react list of goods

Arsen-Shpak:
[DEMO LINK](https://Arsen-Shpak.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -34,62 +35,110 @@ export function getReorderedGoods(
...
+  sortByAlphabet = () => {
+    this.setState({ sortType: SortType.ALPABET });
+  };
+
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  reverseGoods = () => {
+    this.setState((state) => ({ isReversed: !state.isReversed }));
```
TheVovchik, line 78: 👍

src/App.tsx
```diff
@@ -34,62 +35,110 @@ export function getReorderedGoods(
...
+      case SortType.LENGTH:
+        return goodA.length - goodB.length;
+      default:
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
```
TheVovchik, line 51: 👍

src/App.tsx
```diff
@@ -34,62 +35,110 @@ export function getReorderedGoods(
...
+  render() {
+    const { isReversed, sortType } = this.state;
+    const sortedGoods = getReorderedGoods(goodsFromServer, this.state);
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={classNames('button is-info',
```
TheVovchik, line 97: as for me better to split them in separate classes

src/App.tsx
```diff
@@ -34,62 +35,110 @@ export function getReorderedGoods(
...
+            type="button"
+            className={classNames('button is-warning',
+              { 'is-light': !isReversed })}
+            onClick={this.reverseGoods}
+          >
+            Reverse
+          </button>
+
+          {(sortType !== SortType.NONE || isReversed)
+            ? (
```
TheVovchik, line 121: better use logical operator && instead of ternary operator
so if first condition is true it will return second one,
and if first is false it will return false which will be equal of null

src/App.tsx
```diff
@@ -34,62 +35,110 @@ export function getReorderedGoods(
...
+
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {sortedGoods.map(good => (
+            <li data-cy="Good" key={good}>
```
TheVovchik, line 136: have had reject while use a name from array, mentor typed that it can be not unique

TheVovchik (approved):
GJ 👍 
Look on comment on 121 line

PR #1212, opened 2022-10-23
list of goods with implemented methods on buttons. Button was moved to separate component

TheVovchik:
[DEMO LINK](https://TheVovchik.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { uuid } from 'uuidv4';
+import classNames from 'classnames';
+import { Button } from './Components/Button';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -21,75 +24,130 @@ enum SortType {
...
+    });
+  };
+
+  render() {
+    const { sortType, isReversed } = this.state;
+
+    const visibleGoods = implementButtonsActions(goodsFromServer,
+      sortType, isReversed);
+
+    const checkOnChanges = isReversed || sortType !== SortType.NONE;
```
naziamoff, line 88: this one is boolean variable but by its name I guess that it is a function
Booleans normally start with `should`, `is` and so on

src/App.tsx
```diff
@@ -21,75 +24,130 @@ enum SortType {
...
-type ReorderOptions = {
-  sortType: SortType,
+type State = {
   isReversed: boolean,
+  sortType: SortType,
 };
 
-// Use this function in the render to prepare goods
-export function getReorderedGoods(
+export function implementButtonsActions(
```
naziamoff, line 32: function should have single responsibility in order to make your code clearer. Think of a way of splitting this function. And you should avoid having a lot of args inside the function

src/App.tsx
```diff
@@ -21,75 +24,130 @@ enum SortType {
...
-          Reset
-        </button>
-      </div>
+export class App extends Component<{}, State> {
+  state = {
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
+
+  setOrderAlphabetical = () => {
```
naziamoff, line 61: Better to keep it simple here
```suggestion
  sortByAlphabet = () => {
```

src/App.tsx
```diff
@@ -21,75 +24,130 @@ enum SortType {
...
+    this.setState({ sortType: SortType.ALPABET });
+  };
+
+  setOrderLength = () => {
```
naziamoff, line 65: Same here
```suggestion
  sortByLength = () => {
```

src/App.tsx
```diff
@@ -21,75 +24,130 @@ enum SortType {
...
+          <ul>
+            {visibleGoods.map(good => {
+              return (
+                <li
+                  data-cy="Good"
+                  key={uuid()}
+                >
+                  {good}
+                </li>
+              );
```
naziamoff, line 146:
```suggestion
            {visibleGoods.map(good => (
              <li
                data-cy="Good"
                key={uuid()}
              >
                {good}
              </li>
```

naziamoff (changes requested):

src/App.tsx
```diff
@@ -21,75 +24,128 @@ enum SortType {
...
+    });
+  };
+
+  render() {
+    const { sortType, isReversed } = this.state;
+
+    const visibleGoods = getSortedGoods(goodsFromServer,
+      sortType, isReversed);
+
+    const isReset = isReversed || sortType !== SortType.NONE;
```
naziamoff, line 88: I think looks clearer this way
```suggestion
    const shouldRenderReset = isReversed || sortType !== SortType.NONE;
```

naziamoff (approved):
GJ! But I encourage you to avoid taking more than 2 arguments in a function - you should separate methods

PR #1215, opened 2022-10-23
commit1

Gintako666:
[DEMO LINK](https://Gintako666.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import className from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -16,80 +17,127 @@ export const goodsFromServer = [
...
 
 type ReorderOptions = {
   sortType: SortType,
   isReversed: boolean,
 };
 
-// Use this function in the render to prepare goods
-export function getReorderedGoods(
+export function GetReorderedGoods(
   goods: string[],
```
stanislav-topikha, line 31: it`s not ok to return jsx like this in separate function, use this function to reorder items and return reordered list, then map it insine of App component

stanislav-topikha (changes requested):

DarkMistyRoom (approved):

PR #1217, opened 2022-10-24
React list of goods

Yuriy-Komolov:
[DEMO LINK](https://Yuriy-Komolov.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,11 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+import { GoodsList } from './components/GoodsList';
+import { Button } from './components/Button';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


igor-1987 (approved):
Well done!

src/App.tsx
```diff
@@ -17,79 +22,125 @@ export const goodsFromServer = [
...
-};
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  sortByAlphabet = () => {
+    this.setState({ sortType: SortType.ALPHABET });
+  };
+
+  reverse = () => {
```
naziamoff, line 79: It actually changes the previous state of reverse, so it is better to write it explicitly
```suggestion
  toggleReverse = () => {
```

src/App.tsx
```diff
@@ -17,79 +22,125 @@ export const goodsFromServer = [
...
+            className={
+              classNames('is-warning', {
+                'is-light': !isReversed,
+              })
+            }
+          >
+            Reverse
+          </Button>
+
+          {(sortType !== SortType.NONE || isReversed) && (
```
naziamoff, line 132: better to create separate variable for that
If you will need to add one more condition later - you will have to find it in your code. Will be much easier if it is stored inside the separate variable

naziamoff (changes requested):

PR #1219, opened 2022-10-24
Implemented a list of goods which you can sort

anton-iskryk:
[DEMO LINK](https://anton-iskryk.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -51,9 +51,6 @@ export function getReorderedGoods(
     visibleGoods.reverse();
   }
 
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
-
   return visibleGoods;
 }
 
@@ -96,6 +93,7 @@ export class App extends Component<{}, State> {
   render() {
     const { isReversed, sortType } = this.state;
     const isResetButtonVisible = sortType !== SortType.NONE || isReversed;
+
     const visibleItems = getReorderedGoods(
       goodsFromServer,
       { sortType, isReversed },


src/App.tsx
```diff
@@ -26,70 +28,137 @@ type ReorderOptions = {
...
-      <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {getReorderedGoods(goodsFromServer, this.state).map(good => (
```
artem-nasadchuk, line 156: for better readability, it is better to put the result of the function into a variable and here map this variable

src/App.tsx
```diff
@@ -26,70 +28,137 @@ type ReorderOptions = {
...
+              <li key={uuidv4()} data-cy="Good">{good}</li>
```
artem-nasadchuk, line 157:
```suggestion
              <li
                key={uuidv4()}
                data-cy="Good"
              >
                {good}
              </li>
```

src/App.tsx
```diff
@@ -26,70 +28,137 @@ type ReorderOptions = {
...
+  resetList = () => {
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    const visibilityResetButton = sortType !== SortType.NONE || isReversed;
```
naziamoff, line 98: You have chosen the right way of naming your variable 👍 
Still not clear enough
`booleans` normally start with `is`, `should`, `can` and so on - this will make your code easier to read

src/App.tsx
```diff
@@ -26,70 +28,137 @@ type ReorderOptions = {
...
+          >
+            Sort by length
+          </button>
+
+          <button
+            type="button"
+            onClick={() => {
+              this.setState({
+                isReversed: !isReversed,
+              });
```
naziamoff, line 132: Better to create a separate method for that

src/App.tsx
naziamoff, line 156: passing state as func props looks wrong to me. If we have more different properties in state - we will pass them to the function even if it doesn't use them.

src/App.tsx
naziamoff: I agree with @artem-nasadchuk here

naziamoff (changes requested):

src/App.tsx
```diff
@@ -26,70 +28,144 @@ type ReorderOptions = {
...
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
```
naziamoff, line 55: do you need to log it here?

src/App.tsx
```diff
@@ -26,70 +28,144 @@ type ReorderOptions = {
...
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    const isResetButtonVisible = sortType !== SortType.NONE || isReversed;
+    const visibleItems = getReorderedGoods(
```
naziamoff, line 99: I think I would have added an empty line here to make it easier to read

naziamoff (approved):
Well done!

PR #1224, opened 2022-10-25
Create List of goods App with sort, reverse and reset buttons

avkovlit:
[DEMO LINK](https://avkovlit.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -35,6 +36,23 @@ export function getReorderedGoods(
   const visibleGoods = [...goods];
 
   // Sort and reverse goods if needed
+  visibleGoods.sort((good1, good2) => {
+    switch (sortType) {
+      case SortType.ALPABET:
+        return good1.localeCompare(good2);
+
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      default:
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -35,61 +36,127 @@ export function getReorderedGoods(
   const visibleGoods = [...goods];
 
   // Sort and reverse goods if needed
```
naziamoff, line 38: remove unnecessary comments and logs

src/App.tsx
```diff
@@ -35,61 +36,127 @@ export function getReorderedGoods(
...
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+
+    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
```
naziamoff, line 98: bad idea to pass state as a props - if we add some variables there - we will pass the props the function does not need. Think of a different way

naziamoff: ideally those are different functions - reverse and sort

src/App.tsx
```diff
@@ -35,61 +36,127 @@ export function getReorderedGoods(
...
+            Sort alphabetically
+          </button>
+
+          <button
+            type="button"
+            className={classNames('button is-success',
+              {
+                'is-light': sortType !== SortType.LENGTH,
+              })}
+            onClick={this.sortByLength}
```
naziamoff, line 120: 👍

src/App.tsx
```diff
@@ -35,61 +36,127 @@ export function getReorderedGoods(
...
+              {
+                'is-light': !isReversed,
+              })}
+            onClick={this.reverseList}
+          >
+            Reverse
+          </button>
+
+          {
+            (sortType !== SortType.NONE || isReversed) && (
```
naziamoff, line 137: better to create separate variable for that. What if we add some parameter of reordering? You will have to look for this piece of code manually. So the variable is good idea

naziamoff (changes requested):

src/App.tsx
```diff
@@ -26,70 +27,136 @@ type ReorderOptions = {
...
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+
+    const visibleGoods = getReorderedGoods(
+      goodsFromServer, { isReversed, sortType },
+    );
+
+    const isReset = sortType !== SortType.NONE || isReversed;
```
naziamoff, line 98: Not clear what is this variable and why do you need it. Think of a different name for it
```suggestion
    const isReset = sortType !== SortType.NONE || isReversed;
```

naziamoff (approved):

PR #1234, opened 2022-10-27
solution

1stIndominus:
[DEMO LINK](https://1stIndominus.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,18 +22,12 @@ type State = {
   sortBy: string,
   resetVisibility: boolean,
   classVisibility: string,
-  isReverseClicked: boolean,
 };
 
-type ButtonsMap = {
-  ALPHABETICALLY: string,
-  BUY_LENGTH: string,
-};
-
-const buttonsMap: ButtonsMap = {
-  ALPHABETICALLY: 'ALPHABETICALLY',
-  BUY_LENGTH: 'BUY_LENGTH',
-};
+enum ButtonsMap {
+  ALPHABETICALLY = 'ALPHABETICALLY',
+  BY_LENGTH = 'BY_LENGTH',
+}
 
 export class App extends React.Component<{}, State> {
   state = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,81 +16,160 @@ export const goodsFromServer = [
...
-// Use this function in the render to prepare goods
-export function getReorderedGoods(
-  goods: string[],
-  { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
+type ButtonsMap = {
+  ALPHABETICALLY: string,
+  BUY_LENGTH: string,
```
polosanya, line 30:
```suggestion
  BY_LENGTH: string,
```
Please fix this typo in all cases
Btw, it is better to use Enum in this case
https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

src/App.tsx
```diff
@@ -15,81 +16,160 @@ export const goodsFromServer = [
...
+      isReversed,
+      sortBy,
+      resetVisibility,
+      classVisibility,
+      isReverseClicked,
+    } = this.state;
+
+    const visibleGoods = [...goods];
+
+    visibleGoods.sort((alph, len) => {
```
polosanya, line 99: Please use more informative names, than `alph` and `len`

src/App.tsx
```diff
@@ -15,81 +16,160 @@ export const goodsFromServer = [
...
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+  reverse = () => {
+    this.setState(state => ({
+      isReversed: !state.isReversed,
+      resetVisibility: true,
+      isReverseClicked: !state.isReverseClicked,
```
polosanya, line 56: Why do you need both `isReversed` and `isReverseClicked`?
You can use only first one

polosanya (changes requested):
Good job! Please pay attention to my comments

polosanya (approved):
Good job!

PR #1246, opened 2022-11-07
Solution

BZOzyurt:
[DEMO LINK](https://BZOzyurt.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -22,8 +23,8 @@ enum SortType {
 }
 
 type ReorderOptions = {
-  sortType: SortType,
-  isReversed: boolean,
+  sortType: SortType;
+  isReversed: boolean;
 };
 
 // Use this function in the render to prepare goods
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -38,58 +39,120 @@ export function getReorderedGoods(
...
+  visibleGoods.sort((goods1, goods2) => {
+    switch (sortType) {
+      case SortType.ALPABET:
+        return goods1.localeCompare(goods2);
+
+      case SortType.LENGTH:
+        return goods1.length - goods2.length;
+
+      default:
+        return 0;
```
Naveed-Shoukat, line 51: return visibleGoods or null instead of return 0

src/App.tsx
```diff
@@ -38,58 +39,120 @@ export function getReorderedGoods(
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
```
Naveed-Shoukat, line 40: remove console.log

Naveed-Shoukat (changes requested):
Please remove console.log, In your switch case return null or visiableGoods instead of returning 0. Also fix the UI as clicking the reset button shows Zero.

src/App.tsx
VinnieJ-2k20: Nope, `0` is fine here, since it's `sort` method's callback, that should return number.

src/App.tsx
```diff
@@ -22,74 +23,131 @@ enum SortType {
...
+          </button>
+
+          {(isReversed || sortType) && (
+            <button
+              type="button"
+              className="button is-danger is-light"
+              onClick={this.reset}
+            >
+              Reset
+            </button>
```
VinnieJ-2k20, line 137: Yep, fix remove shower `0` on `UI` of Reset is inactive
image

VinnieJ-2k20 (changes requested):
One thing to fix on UI. Almost done.

VinnieJ-2k20 (approved):

PR #1251, opened 2022-11-08
Solution for list of goods

sapnachoudhary06:
- [DEMO LINK](https://sapnachoudhary06.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -34,62 +35,136 @@ export function getReorderedGoods(
...
+      break;
 
-  return visibleGoods;
+    case SortType.LENGTH:
+      visibleGoods.sort((good1, good2) => (
+        good1.length - good2.length
+      ));
+      break;
+
+    default:
```
Naveed-Shoukat, line 51: you may consider to return visibleGoods as a default option.

Naveed-Shoukat (approved):

src/App.tsx
VinnieJ-2k20, line 51: Try using switch/case inside sort callback.

VinnieJ-2k20 (approved):
Looks good. Added one comment.

PR #1279, opened 2022-11-24
done

yaro-bey:
[DEMO LINK](https://yaro-bey.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,70 +27,135 @@ type ReorderOptions = {
...
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            onClick={this.sortAlphabetically}
+            className={classNames(
+              'button',
+              'is-info',
```
Anatolii-Voloshyn, line 101: you can use it like one class 'button is-info' and add only 'is-light', like second parameter

Anatolii-Voloshyn (approved):
GJ!

src/App.tsx
```diff
@@ -26,70 +27,133 @@ type ReorderOptions = {
...
+  reset = () => {
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
```
denys-danyliuk, line 91:
```suggestion
    const { isReversed, sortType } = this.state;
    
    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
```
add space between destructuring and regular variables for better readability

src/App.tsx
denys-danyliuk: agree

src/App.tsx
```diff
@@ -26,70 +27,133 @@ type ReorderOptions = {
...
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            onClick={this.sortAlphabetically}
+            className={classNames(
+              'button',
+              'is-info',
```
denys-danyliuk, line 101:
```suggestion
              'button, is-info',
```

src/App.tsx
```diff
@@ -26,70 +27,133 @@ type ReorderOptions = {
...
+              'button is-warning',
+              {
+                'is-light': !isReversed,
+              },
+            )}
+          >
+            Reverse
+          </button>
+
+          {(isReversed || sortType !== SortType.NONE) && (
```
denys-danyliuk, line 136: can move this condition to variable

denys-danyliuk (approved):

PR #1286, opened 2022-11-25
list of goods

Svitlana-Yudina:
[DEMO LINK](https://Svitlana-Yudina.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,7 +37,23 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
+  visibleGoods.sort((good1, good2) => {
+    switch (sortType) {
+      case SortType.ALPABET:
+        return good1.localeCompare(good2);
+
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      default:
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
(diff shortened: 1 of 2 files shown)


mitattt (approved):
good job

sh1gatsu (changes requested):
image

`wrong color of buttons in Demo link`

src/App.tsx
```diff
@@ -34,62 +35,127 @@ export function getReorderedGoods(
...
-  // Sort and reverse goods if needed
+  visibleGoods.sort((good1, good2) => {
+    switch (sortType) {
+      case SortType.ALPABET:
+        return good1.localeCompare(good2);
+
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      default:
```
denys-danyliuk, line 46:
```suggestion
      case SortType.NONE
```
add all known cases

src/App.tsx
```diff
@@ -34,62 +35,127 @@ export function getReorderedGoods(
...
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
+
+  render() {
+    const { sortType, isReversed } = this.state;
+    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
+    const isOrderChanged = isReversed || sortType !== SortType.NONE;
```
denys-danyliuk, line 97:
```suggestion
    const { sortType, isReversed } = this.state;
    
    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
    const isOrderChanged = isReversed || sortType !== SortType.NONE;
```
it's better to separate destructuring and regular variables for better readability

src/App.tsx
```diff
@@ -34,62 +35,127 @@ export function getReorderedGoods(
...
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={classNames(
+              'button',
+              'is-info',
+              { 'is-light': sortType !== SortType.ALPABET },
```
denys-danyliuk, line 107:
```suggestion
              { 
                'is-light': sortType !== SortType.ALPABET 
              },
```

denys-danyliuk (changes requested):

src/App.tsx
Svitlana-Yudina: I'll add. But I don't understand why? I have cases in which I do different actions ... Why is it necessary to list everything if I will do in these cases what I do by default?

mykola-koval: cases should cover 100% of actions that you can pass in `sortType`. If you will not do this and no case works out, your switch returns `undefined` and that can cause bugs.

mykola-koval (approved):

PR #1290, opened 2022-11-25
added solution

kryvets-vadym:
[DEMO LINK](https://kryvets-vadym.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,7 +18,7 @@ export const goodsFromServer = [
 
 enum SortType {
   NONE,
-  ALPABET,
+  ALPHABET,
   LENGTH,
 }
 
@@ -35,7 +35,7 @@ export function getReorderedGoods(
 
   visibleGoods.sort((firstGood, secondGood) => {
     switch (sortType) {
-      case SortType.ALPABET:
+      case SortType.ALPHABET:
         return firstGood.localeCompare(secondGood);
 
       case SortType.LENGTH:
@@ -66,7 +66,7 @@ export class App extends Component<{}, State> {
   };
 
   sortByAlphabet = () => {
-    this.setState({ sortType: SortType.ALPABET });
+    this.setState({ sortType: SortType.ALPHABET });
   };
 
   sortByLength = () => {
@@ -103,7 +103,7 @@ export class App extends Component<{}, State> {
           <button
             type="button"
             className={classNames('button', 'is-info', {
-              'is-light': sortType !== SortType.ALPABET,
+              'is-light': sortType !== SortType.ALPHABET,
             })}
             onClick={sortByAlphabet}
           >
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -93,6 +93,7 @@ export class App extends Component<{}, State> {
   render() {
     const { sortType, isReversed } = this.state;
     const { reverse, setSortType, reset } = this;
+    const isOrderChanged = isReversed || sortType !== SortType.NONE;
     const orderedGoods = getReorderedGoods(goodsFromServer, this.state);
 
     return (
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +28,127 @@ type ReorderOptions = {
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {orderedGoods.map(good => (
+              <li data-cy="Good">{ good }</li>
+            ))}
```
Stepico, line 148: I think the key attribute is must-have

src/App.tsx
```diff
@@ -26,70 +28,127 @@ type ReorderOptions = {
...
+              'is-light': !isReversed,
+            })}
+            onClick={() => reverse()}
+          >
+            Reverse
+          </button>
+
+          {sortType === SortType.NONE && !isReversed
+            ? ('')
+            : (
```
Stepico, line 133: I don't think that you need ternary here because, use conditional rendering

Stepico (changes requested):
Seems good to me, just some minor suggestions from me

Stepico (approved):
GJ, also I noticed that you can click on an active button and it becomes inactive and it works with Sort buttons, but the mockup doesn't show it, yet it's better that you added some additional functionality, but reset button in this way is useless I guess, so I'm not sure about this one

src/App.tsx
```diff
@@ -26,70 +28,126 @@ type ReorderOptions = {
...
+  setSortType = (sortType: SortType) => {
+    return this.state.sortType === sortType
+      ? this.setState({
+        sortType: SortType.NONE,
+      })
+
+      : this.setState({
+        sortType,
+      });
+  };
```
mykola-koval, line 84: I guess we should simplify this or even create a function for each type of sorting

mykola-koval: right now i dont see any benefits in using this function over regular setState

src/App.tsx
```diff
@@ -26,70 +28,126 @@ type ReorderOptions = {
...
+            onClick={() => reverse()}
+          >
+            Reverse
+          </button>
+
+          {isOrderChanged && (
+            <button
+              type="button"
+              className="button is-danger is-light"
+              onClick={() => reset()}
```
mykola-koval, line 136:
```suggestion
              onClick={reset}
```

src/App.tsx
```diff
@@ -26,70 +28,126 @@ type ReorderOptions = {
...
+          >
+            Sort by length
+          </button>
+
+          <button
+            type="button"
+            className={classNames('button', 'is-warning', {
+              'is-light': !isReversed,
+            })}
+            onClick={() => reverse()}
```
mykola-koval, line 127:
```suggestion
            onClick={reverse}
```

src/App.tsx
```diff
@@ -26,70 +28,126 @@ type ReorderOptions = {
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {orderedGoods.map(good => (
+              <li data-cy="Good" key={good}>{ good }</li>
```
mykola-koval, line 146:
```suggestion
              <li data-cy="Good" key={good}>
                {good}
              </li>
```

src/App.tsx
```diff
@@ -26,70 +27,128 @@ type ReorderOptions = {
...
+  render() {
+    const { sortType, isReversed } = this.state;
+    const {
+      sortByAlphabet,
+      sortByLength,
+      reverse,
+      reset,
+    } = this;
+    const isOrderChanged = isReversed || sortType !== SortType.NONE;
+    const orderedGoods = getReorderedGoods(goodsFromServer, this.state);
```
denys-danyliuk, line 98:
```suggestion
    const { sortType, isReversed } = this.state;
    const {
      sortByAlphabet,
      sortByLength,
      reverse,
      reset,
    } = this;
    
    const isOrderChanged = isReversed || sortType !== SortType.NONE;
    
    const orderedGoods = getReorderedGoods(goodsFromServer, this.state);
```

src/App.tsx
```diff
@@ -26,70 +27,128 @@ type ReorderOptions = {
...
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={classNames('button', 'is-info', {
+              'is-light': sortType !== SortType.ALPHABET,
+            })}
```
denys-danyliuk, line 107:
```suggestion
            className={classNames('button', 'is-info', 
            {
              'is-light': sortType !== SortType.ALPHABET,
            }
          )}
```

src/App.tsx
denys-danyliuk: fix it

denys-danyliuk (approved):

PR #1313, opened 2022-12-03
List of goods task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -33,16 +33,16 @@ export function getReorderedGoods(
   const visibleGoods = [...goods];
 
   switch (sortType) {
-    case 0:
+    case SortType.NONE:
       break;
 
-    case 1:
+    case SortType.ALPABET:
       visibleGoods.sort(
         (good1, good2) => good1.localeCompare(good2),
       );
       break;
 
-    case 2:
+    case SortType.LENGTH:
       visibleGoods.sort(
         (good1, good2) => (good1.length - good2.length),
       );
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +26,135 @@ type ReorderOptions = {
...
+    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
+    const { isReversed, sortType, isReset } = this.state;
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={
+              `button is-info ${(sortType === SortType.ALPHABET) ? '' : 'is-light'}`
```
TarasHoliuk, line 114: it would look better if you will use `classnames`

src/App.tsx
```diff
@@ -26,70 +26,135 @@ type ReorderOptions = {
...
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+  isReset: boolean,
```
TarasHoliuk, line 65: the same as for variant with hooks - you can calculate it from 2 previous state variables

TarasHoliuk (changes requested):

src/App.tsx
juliiaap: fixed

src/App.tsx
juliiaap: fixed

loralevitska (approved):
good job!

PR #1321, opened 2022-12-10
Solution

berko228:
[DEMO LINK](https://berko228.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,70 +27,120 @@ type ReorderOptions = {
...
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((g1, g2) => {
+    switch (sortType) {
+      case 1:
```
TarasHoliuk, line 38:
```suggestion
      case SortType.ALPHABET:
```
It's unclear to use numbers from Enum. Use it as values of object instead :arrow_up:

src/App.tsx
```diff
@@ -26,70 +27,120 @@ type ReorderOptions = {
...
+    this.setState((state) => ({
+      isReversed: !state.isReversed,
+    }));
+  };
+
+  render() {
+    const currentGoods = getReorderedGoods(goodsFromServer, this.state);
+    const { sortType, isReversed } = this.state;
+    const resetCondition
+      = (sortType !== 0) || (sortType === 0 && isReversed === true);
```
TarasHoliuk, line 75:
```suggestion
    const resetCondition = sortType !== SortType.NONE || isReversed;
```
If it's `reversed` we want to show `Reset` button anyway - doesn't matter which sort type is chosen

TarasHoliuk (changes requested):
GJ! I've added 2 small comments

berko228:
> fix all comments, please
> 
> image
> 
> 

Already did it

PR #1326, opened 2022-12-15
Soluion

UmizDemud:
[DEMO LINK](https://UmizDemud.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -34,62 +34,88 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  if (sortType === 1) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  } else if (sortType === 2) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  }
 
-  return visibleGoods;
+  return isReversed
+    ? visibleGoods.reverse()
+    : visibleGoods;
 }
 
 // DON'T save goods to the state
-// type State = {
-//   isReversed: boolean,
-//   sortType: SortType,
-// };
-
-export const App: React.FC = () => {
-  return (
-    <div className="section content">
-      <div className="buttons">
-        <button
-          type="button"
-          className="button is-info is-light"
-        >
-          Sort alphabetically
-        </button>
-
-        <button
-          type="button"
-          className="button is-success is-light"
-        >
-          Sort by length
-        </button>
-
-        <button
-          type="button"
-          className="button is-warning is-light"
-        >
-          Reverse
-        </button>
-
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
-
-      <ul>
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
-      </ul>
-    </div>
-  );
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
 };
+
+export class App extends React.Component<{}, State> {
+  state = {
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
+
+  render() {
+    const goods = getReorderedGoods(
+      goodsFromServer,
+      this.state,
+    );
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={`button is-info${this.state.sortType !== 1 ? ' is-light' : ''}`}
+            onClick={() => this.setState({ sortType: 1 })}
+          >
+            Sort alphabetically
+          </button>
+
+          <button
+            type="button"
+            className={`button is-success${this.state.sortType !== 2 ? ' is-light' : ''}`}
+            onClick={() => this.setState({ sortType: 2 })}
+          >
+            Sort by length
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -34,62 +34,88 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  if (sortType === 1) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  } else if (sortType === 2) {
```
meforoma, line 39: `=== number` is not a good idea when you have an enum
apply this for the whole code

```suggestion
  if (sortType === SortType.ALPHABET) {
    visibleGoods.sort((a, b) => a.localeCompare(b));
  } else if (sortType === SortType.LENGTH) {
```

src/App.tsx
```diff
@@ -34,62 +34,88 @@ export function getReorderedGoods(
...
-        >
-          Reverse
-        </button>
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={`button is-info${this.state.sortType !== 1 ? ' is-light' : ''}`}
+            onClick={() => this.setState({ sortType: 1 })}
```
meforoma, line 72: use enum

src/App.tsx
```diff
@@ -34,62 +34,88 @@ export function getReorderedGoods(
...
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+          <button
+            type="button"
+            className={`button is-success${this.state.sortType !== 2 ? ' is-light' : ''}`}
+            onClick={() => this.setState({ sortType: 2 })}
```
meforoma, line 80: use enum

src/App.tsx
```diff
@@ -34,62 +34,88 @@ export function getReorderedGoods(
...
+                className="button is-danger is-light"
+                onClick={() => this.setState({
+                  isReversed: false,
+                  sortType: 0,
+                })}
+              >
+                Reset
+              </button>
+            )
+            : <></>}
```
meforoma, line 108: better use null instead of an empty fragment
```suggestion
            : null}
```

meforoma, line 108: and overall it's even easier - just conditionally render the whole button - without ternary, but with `&&`
```suggestion
          {this.state.sortType !== 0 || this.state.isReversed
            && (
              <button
                type="button"
                className="button is-danger is-light"
                onClick={() => this.setState({
                  isReversed: false,
                  sortType: 0,
                })}
              >
                Reset
              </button>
            )}
```

src/App.tsx
```diff
@@ -34,62 +34,88 @@ export function getReorderedGoods(
...
+          </button>
+
+          {this.state.sortType !== 0 || this.state.isReversed
+            ? (
+              <button
+                type="button"
+                className="button is-danger is-light"
+                onClick={() => this.setState({
+                  isReversed: false,
+                  sortType: 0,
```
meforoma, line 102: enum of course )

src/App.tsx
```diff
@@ -34,62 +34,88 @@ export function getReorderedGoods(
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {goods.map((item => (
+              <li key={item} data-cy="Good">{item}</li>
```
meforoma, line 114: hmm, setting an entire object as a key?
does the item have an `id`? If it is, i'd rather go with `item.id`
```suggestion
              <li key={item.id} data-cy="Good">{item}</li>
```

meforoma (changes requested):
please see the comments to address

src/App.tsx
UmizDemud: No this is just an array of strings

```js
export const goodsFromServer = [
  'Dumplings',
  'Carrot',
  'Eggs',
  'Ice cream',
  'Apple',
  'Bread',
  'Fish',
  'Honey',
  'Jam',
  'Garlic',
];
```

src/App.tsx
UmizDemud: I memorised the numbers 😄

src/App.tsx
UmizDemud: Sometimes I look back at my code, and ask myself why..

Looks like because I take breaks very rarely

UmizDemud: 👍 
just realized this ( 99% would be faster )

meforoma (changes requested):
i can't see the code changes according to comments @UmizDemud 
could you forget to commit & push?
Or you just don't want to apply changes? 🙂

UmizDemud:
> i can't see the code changes according to comments @UmizDemud could you forget to commit & push? Or you just don't want to apply changes? 🙂

Thanks for noticing. Turns out I improved the react_list-of-goods-hooks task :D

src/App.tsx
```diff
@@ -63,49 +63,53 @@ export class App extends React.Component<{}, State> {
...
+            onClick={() => this.setState({ sortType: SortType.ALPHABET })}
           >
             Sort alphabetically
           </button>
 
           <button
             type="button"
-            className={`button is-success${this.state.sortType !== 2 ? ' is-light' : ''}`}
-            onClick={() => this.setState({ sortType: 2 })}
+            className={`button is-success${sortType !== SortType.LENGTH ? ' is-light' : ''}`}
```
meforoma, line 84: you could also used the `classnames` package for conditional classes assignment:
https://www.npmjs.com/package/classnames

it's pretty useful - in Mate we use it super often

meforoma (approved):
🎉

PR #1327, opened 2022-12-15
finished

metinbicaksiz:
[DEMO LINK](https://metinbicaksiz.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,75 +21,139 @@ enum SortType {
   LENGTH,
 }
 
-type ReorderOptions = {
-  sortType: SortType,
-  isReversed: boolean,
-};
-
-// Use this function in the render to prepare goods
-export function getReorderedGoods(
-  goods: string[],
-  { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
-
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
-
-  return visibleGoods;
-}
-
-// DON'T save goods to the state
-// type State = {
-//   isReversed: boolean,
+// type ReorderOptions = {
 //   sortType: SortType,
+//   isReversed: boolean,
 // };
 
-export const App: React.FC = () => {
-  return (
-    <div className="section content">
-      <div className="buttons">
-        <button
-          type="button"
-          className="button is-info is-light"
-        >
-          Sort alphabetically
-        </button>
+// Use this function in the render to prepare goods
+// export function getReorderedGoods(
+//   goods: string[],
+//   { sortType, isReversed }: ReorderOptions,
+// ) {
+//   // To avoid the original array mutation
+//   const visibleGoods = [...goods];
 
-        <button
-          type="button"
-          className="button is-success is-light"
-        >
-          Sort by length
-        </button>
+//   // Sort and reverse goods if needed
+//   // eslint-disable-next-line no-console
+//   console.log(sortType, isReversed);
 
-        <button
-          type="button"
-          className="button is-warning is-light"
-        >
-          Reverse
-        </button>
+//   return visibleGoods;
+// }
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
-
-      <ul>
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -21,75 +21,139 @@ enum SortType {
...
 
 // Use this function in the render to prepare goods
-export function getReorderedGoods(
-  goods: string[],
-  { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
+// export function getReorderedGoods(
+//   goods: string[],
```
polosanya, line 31: Please delete all the comments that you don't need

src/App.tsx
```diff
@@ -21,75 +21,139 @@ enum SortType {
...
+      visibleGoods.reverse();
+    }
+
+    return (
+      <>
+        <div className="section content">
+          <div className="buttons">
+            <button
+              type="button"
+              className={`button is-info ${this.state.sortType !== SortType.ALPHABET ? 'is-light' : ''}`}
```
polosanya, line 85: You can also use [classnames too](https://www.npmjs.com/package/classnames)l for that

src/App.tsx
```diff
@@ -21,75 +21,139 @@ enum SortType {
...
+              onClick={
+                () => {
```
polosanya, line 87: Let's move this logic into a separate function to make your code cleaner

src/App.tsx
```diff
@@ -21,75 +21,139 @@ enum SortType {
...
+              }
+            >
+              Sort alphabetically
+            </button>
+
+            <button
+              type="button"
+              className={`button is-success ${this.state.sortType !== SortType.LENGTH ? 'is-light' : ''}`}
+              onClick={
+                (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
```
polosanya, line 101: Same here

src/App.tsx
```diff
@@ -21,75 +21,139 @@ enum SortType {
...
+              }
+            >
+              Sort by length
+            </button>
+
+            <button
+              type="button"
+              className={`button is-warning ${this.state.isReversed !== true ? 'is-light' : ''}`}
+              onClick={
+                () => {
```
polosanya, line 116: And here

src/App.tsx
```diff
@@ -21,75 +21,139 @@ enum SortType {
...
+            >
+              Reverse
+            </button>
+
+            {
+              (visibleGoods[0] !== this.state.goods[0]) && (
+                <button
+                  type="button"
+                  className="button is-danger is-light"
+                  onClick={() => {
```
polosanya, line 131: Also here. Please fix all the cases

src/App.tsx
```diff
@@ -21,75 +21,139 @@ enum SortType {
...
+                  }}
+                >
+                  Reset
+                </button>
+              )
+            }
+          </div>
+
+          <ul>
+            <ul>
```
polosanya, line 145: Do you really need to have two nested `ul` tags here?

polosanya (changes requested):
Almost perfect! Please check my comments

src/App.tsx
jstmpelowycz: fix it, please

src/App.tsx
jstmpelowycz: fix it, please

src/App.tsx
jstmpelowycz: and this one, too

src/App.tsx
jstmpelowycz: and here

src/App.tsx
```diff
@@ -21,75 +21,139 @@ enum SortType {
...
+                    isReversed: !state.isReversed,
+                  }));
+                }
+              }
+            >
+              Reverse
+            </button>
+
+            {
+              (visibleGoods[0] !== this.state.goods[0]) && (
```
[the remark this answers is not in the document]
jstmpelowycz: I do think so as well, however – up to you

src/App.tsx
jstmpelowycz: and this one as well

src/App.tsx
jstmpelowycz: yes, you don't have to persist both, kill one of these

jstmpelowycz (changes requested):

src/App.tsx
jstmpelowycz: what about this one?)

jstmpelowycz (approved):
good job!

polosanya (approved):
You do not need to re-request review again, because this task is already approved 🙂

PR #1335, opened 2022-12-25
react_list-of-goods

ysrckr:
[DEMO LINK](https://ysrckr.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
@@ -22,8 +22,8 @@ enum SortType {
 }
 
 type ReorderOptions = {
-  sortType: SortType,
-  isReversed: boolean,
+  sortType: SortType;
+  isReversed: boolean;
 };
 
 // Use this function in the render to prepare goods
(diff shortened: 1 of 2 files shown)


UmizDemud (approved):
I wonder how would it work in these either/or cases with classnames library. I feel the way you did it is bound to be faster. 

👍 good job all around

src/App.tsx
```diff
@@ -48,46 +62,86 @@ export function getReorderedGoods(
...
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={
+            sortType === SortType.ALPHABET
+              ? 'button is-info'
+              : 'button is-info is-light'
+          }
```
naziamoff, line 99: classnames library

src/App.tsx
```diff
@@ -48,46 +62,86 @@ export function getReorderedGoods(
...
+          onClick={sortAlphabetically}
         >
           Sort alphabetically
         </button>
 
         <button
           type="button"
-          className="button is-success is-light"
+          className={
```
naziamoff, line 107: classnames

src/App.tsx
```diff
@@ -48,46 +62,86 @@ export function getReorderedGoods(
...
+          }
+          onClick={sortByLength}
         >
           Sort by length
         </button>
 
         <button
           type="button"
-          className="button is-warning is-light"
+          className={
```
naziamoff, line 119: and here

src/App.tsx
```diff
@@ -48,46 +62,86 @@ export function getReorderedGoods(
...
           Reverse
         </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
+        {(sortType !== SortType.NONE || isReversed) && (
```
naziamoff, line 127: better to make this condition a separate variable. If you decide to add sth to it later - you will need to look for this place. If inside a variable - will be pretty obvious where to add the condition

naziamoff (changes requested):
Not bad at all 🔥 
But still a great reason to practice classnames library)

src/App.tsx
UmizDemud: Now I see. I was working late at night, not the best idea 😃

naziamoff (approved):
Well done 🚀

PR #1336, opened 2022-12-29
Solution react_list_of_goods

anton-martyniuk1:
[DEMO LINK](https://anton-martyniuk1.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,75 +21,110 @@ enum SortType {
   LENGTH,
 }
 
-type ReorderOptions = {
+type State = {
   sortType: SortType,
   isReversed: boolean,
 };
 
-// Use this function in the render to prepare goods
-export function getReorderedGoods(
-  goods: string[],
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -21,75 +21,110 @@ enum SortType {
...
-// Use this function in the render to prepare goods
-export function getReorderedGoods(
-  goods: string[],
-  { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
+export class App extends React.Component<{}, State> {
+  state = {
+    sortType: 0,
```
polosanya, line 31: You have an Enum with sort types. Use it, why do you need to hardcore it with numbers?

polosanya (changes requested):
Almost done! Let's use Enum to improve your code

loralevitska (approved):
😎

PR #1337, opened 2022-12-29
done

Alina-Yermolenko:
[DEMO LINK](https://Alina-Yermolenko.github.io/react_list-of-goods/)


src/App.tsx
```diff
@@ -26,70 +27,126 @@ type ReorderOptions = {
...
 };
 
-// Use this function in the render to prepare goods
-export function getReorderedGoods(
-  goods: string[],
-  { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
+// DON'T save goods to the state
```
polosanya, line 30: Remove useless comments

src/App.tsx
```diff
@@ -26,70 +27,126 @@ type ReorderOptions = {
...
-        </button>
-
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+    if (sortType === 1) {
```
polosanya, line 48: Use switch case to iterate through all possible sortTypes. 
Also, why do you hardcore the values?
You an Enum with sort types, use it

src/App.tsx
```diff
@@ -26,70 +27,126 @@ type ReorderOptions = {
...
+          <button
+            type="button"
+            className={`button is-warning ${!this.state.isReversed ? 'is-light' : ''}`}
+            onClick={this.reverse}
+          >
+            Reverse
+          </button>
+          {
+            JSON.stringify(reorderedGoods)
+              !== JSON.stringify(goodsFromServer)
```
polosanya, line 125: Interesting solution. But you can also check if sort type is none and isReversed is false.

src/App.tsx
```diff
@@ -26,70 +27,126 @@ type ReorderOptions = {
...
+              ? (
+                <button
+                  type="button"
+                  className="button is-danger is-light"
+                  onClick={this.reset}
+                >
+                  Reset
+                </button>
+              ) : <></>
```
polosanya, line 134: No need to return an empty tag. Just use && like this:
`true && <button>...</button>`

polosanya (changes requested):
Almost done! Please check the comments to improve your code a bit)

qoa11a (approved):

PR #1348, opened 2023-01-03
Develop

halynarydasova:
[DEMO LINK](https://halynarydasova.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -39,7 +39,9 @@ export function getReorderedGoods(
   console.log(sortType, isReversed);
 
   switch (sortType) {
-    case SortType.ALPHABET: visibleGoods.sort(); break;
+    case SortType.ALPHABET: visibleGoods.sort(
+      (a, b) => a.localeCompare(b),
+    ); break;
 
     case SortType.LENGTH: visibleGoods.sort(
       (a, b) => a.length - b.length,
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -38,58 +38,110 @@ export function getReorderedGoods(
...
   console.log(sortType, isReversed);
 
+  switch (sortType) {
+    case SortType.ALPHABET: visibleGoods.sort(
+      (a, b) => a.localeCompare(b),
+    ); break;
+
+    case SortType.LENGTH: visibleGoods.sort(
+      (a, b) => a.length - b.length,
+    ); break;
```
jstmpelowycz, line 48:
```suggestion
    case SortType.ALPHABET: 
      visibleGoods.sort((a, b) => a.localeCompare(b));
      break;

    case SortType.LENGTH: 
      visibleGoods.sort((a, b) => a.length - b.length); 
      break;
```

src/App.tsx
```diff
@@ -38,58 +38,110 @@ export function getReorderedGoods(
...
+            onClick={this.handleAlphabeticSort}
+          >
+            Sort alphabetically
+          </button>
+
+          <button
+            type="button"
+            className={`button is-success ${sortType !== SortType.LENGTH
+              ? 'is-light'
+              : ''}`}
```
jstmpelowycz, line 112: suggestion: I'd rather use [classnames](https://www.npmjs.com/package/classnames) library here (I use it often when writing real commercial stuff)

src/App.tsx
```diff
@@ -38,58 +38,110 @@ export function getReorderedGoods(
...
+            onClick={this.handleLengthSort}
+          >
+            Sort by length
+          </button>
+
+          <button
+            type="button"
+            className={`button is-warning ${!isReversed ? 'is-light' : ''}`}
```
jstmpelowycz, line 120: the same here about classnames, just a suggestion, up to you

jstmpelowycz (changes requested):
awesome! almost done

qoa11a (approved):

PR #1350, opened 2023-01-03
Solution

nurdanokcu:
[DEMO LINK](https://nurdanokcu.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +28,145 @@ type ReorderOptions = {
...
+  isReversed: boolean,
+  sortType: SortType,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    goods: goodsFromServer,
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
```
IoghiStudio, line 73: We should not add goods to state, sorry for request changes.

IoghiStudio (changes requested):

src/App.tsx
nurdanokcu: Thank you for the review. I removed it. You can give it another look.

IoghiStudio: Awesome!

IoghiStudio (approved):
Awesome!

PR #1368, opened 2023-01-04
solution

batstolya:
[DEMO LINK](https://batstolya.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,7 +15,6 @@ export const goodsFromServer = [
   'Jam',
   'Garlic',
 ];
-
 enum SortType {
   NONE,
   ALPHABET,
@@ -33,18 +32,21 @@ export function getReorderedGoods(
 ) {
   const visibleGoods = [...goods];
 
-  visibleGoods.sort((a, b) => {
-    switch (sortType) {
-      case SortType.ALPHABET:
-        return a.localeCompare(b);
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => (
+        a.localeCompare(b)));
+      break;
 
-      case SortType.LENGTH:
-        return a.length - b.length;
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => (
+        a.length - b.length));
+      break;
 
-      default:
-        return 0;
-    }
-  });
+    case SortType.NONE:
+    default:
+      break;
+  }
 
   if (isReversed) {
     visibleGoods.reverse();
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -22,74 +23,125 @@ enum SortType {
...
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((a, b) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return a.localeCompare(b);
```
Sveta-Kryukova, line 39: you shouldn't sort it before sortType

    switch (sortType) {
      case SortType.ALPHABET:
        visibleGoods.sort((a, b) => {...

Sveta-Kryukova (changes requested):

Sveta-Kryukova:
and also your styles doesn't work unfortunatly(

src/App.tsx
batstolya: good point, thanks

Sveta-Kryukova (approved):
Good job)

src/App.tsx
```diff
@@ -14,82 +15,137 @@ export const goodsFromServer = [
...
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
+
+  handleClickSortByAlphabetically = () => {
+    this.setState(state => {
+      return { ...state, sortType: SortType.ALPHABET };
```
solaryasha, line 71: Why do you need functional setState here ?

src/App.tsx
```diff
@@ -14,82 +15,137 @@ export const goodsFromServer = [
...
+    });
+  };
+
+  handleClickSortByLength = () => {
+    this.setState(state => {
+      return { ...state, sortType: SortType.LENGTH };
```
solaryasha, line 77: Why do you need functional setState here ?

src/App.tsx
```diff
@@ -14,82 +15,137 @@ export const goodsFromServer = [
...
+    });
+  };
+
+  handleClickReverse = () => {
+    this.setState(state => ({ ...state, isReversed: !state.isReversed }));
```
solaryasha, line 82: Why do you need functional setState here ?

solaryasha (changes requested):

src/App.tsx
batstolya: @solaryasha  In this case i believe, i need functional setState because i use previous value

src/App.tsx
batstolya: Good point! i don't need to use func SetState.

valerii-derkach (approved):

PR #1374, opened 2023-01-05
Add solution

VladimirEdelmann:
# Task
Implement the ability to reorder the goods with the given buttons.
1. `Sort alphabetically` button should sort the goods alphabetically.
    - remove `is-light` class from the button when this order is applied;
1. `Sort by length` button should sort the goods by length.
    - remove `is-light` class from the button when this order is applied;
1. `Reverse` button changes the current order to the opposite one. 
    - it should work together with any sorting;
    - remove `is-light` class from the button when goods are reversed;
    - the second click should show goods in the direct order;
1. `Reset` button should show the goods in the initial order.
    - it should be visible only if goods are not already in the original order;

# Implementation
[DEMO LINK](https://vladimiredelmann.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
@@ -26,7 +27,6 @@ type ReorderOptions = {
   isReversed: boolean,
 };
 
-// Use this function in the render to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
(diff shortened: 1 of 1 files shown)


DiDemkiv (approved):
good job 🚀

src/App.tsx
```diff
@@ -26,70 +27,129 @@ type ReorderOptions = {
...
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
+  // Sort goods if needed
+  if (sortType === SortType.ALPHABET) {
```
solaryasha, line 39: Let's practice `switch` instruction

src/App.tsx
```diff
@@ -26,70 +27,129 @@ type ReorderOptions = {
...
+            type="button"
+            className={
+              classNames(
+                'button is-success',
+                { 'is-light': sortType !== SortType.LENGTH },
+              )
+            }
+            onClick={() => this.setState({
+              sortType: SortType.LENGTH,
+            })}
```
solaryasha, line 109: Don't use inline callbacks, we don't need to create a new arrow function for each re-render

solaryasha: Create class methods

solaryasha (changes requested):

valerii-derkach (changes requested):

valerii-derkach (approved):

src/App.tsx
```diff
@@ -26,70 +27,144 @@ type ReorderOptions = {
...
+      );
+      break;
+
+    case SortType.LENGTH:
+      visibleGoods.sort(
+        (firstGood, secondGood) => firstGood.length - secondGood.length,
+      );
+      break;
+
+    default:
```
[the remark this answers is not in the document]
valerii-derkach: it is better to handle all cases from enum to avoid unexpected behavior in the future, - what if another developer passes this unusable value to your function, and forgets to check if this case is handled? So best practice is to handle this case also and just break, and do not leave space for mistakes

PR #1383, opened 2023-01-05
fisrt commit, with too bad styling due to blackout

vladimir-arkanov:
[DEMO LINK](https://vladimir-arkanov.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import Box from '@mui/material/Box';
+import Paper from '@mui/material/Paper';
+import Button from '@mui/material/Button';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -16,80 +19,171 @@ export const goodsFromServer = [
...
+      break;
+
+    case SortType.LENGTH:
+      visibleGoods.sort((product1, product2) => (
+        product1.length - product2.length
+      ));
+      break;
+
+    default:
+      break;
```
Sveta-Kryukova, line 53: I think we can add that NONE type will be default)

case SortType.NONE:
default:
break;

src/App.tsx
```diff
@@ -16,80 +19,171 @@ export const goodsFromServer = [
...
+            >
+              <div className="section content">
+                <div className="buttons">
+                  <Button
+                    variant="contained"
+                    type="button"
+                    className={sortType === SortType.ALPHABET
+                      ? 'button is-info'
+                      : 'button is-info is-light'}
+                    onClick={this.sortByAlphabet}
```
Sveta-Kryukova, line 131: you can also do it with imported classNames, I think it can save your time) but your solution also good as for me)

src/App.tsx
```diff
@@ -16,80 +19,171 @@ export const goodsFromServer = [
...
+                        variant="outlined"
+                        color="error"
+                        type="button"
+                        className="button is-danger"
+                        onClick={this.resetButton}
+                      >
+                        Reset
+                      </Button>
+                    )
+                    : ''}
```
Sveta-Kryukova, line 170: you shouldn't render this part if it is not on a page) Try to use 

{sortType !== SortType.NONE || isReversed && (...

Sveta-Kryukova (changes requested):

src/App.tsx
vladimir-arkanov: Thanks for great advice, i didn't think about this!

ivan-kadykalo (approved):
Good job 🔥

PR #1385, opened 2023-01-05
React list of goods

OlexandraKorol:
[DEMO LINK](https://OlexandraKorol.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -35,20 +35,22 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((good1: string, good2: string) => {
+        return good1.localeCompare(good2);
+      })
+      break;
 
-  if (sortType !== SortType.NONE) {
-    visibleGoods.sort((good1: string, good2: string) => {
-      switch (sortType) {
-        case SortType.ALPHABET:
-          return good1.localeCompare(good2);
+    case SortType.LENGTH:
+      visibleGoods.sort((good1: string, good2: string) => {
+        return good1.length - good2.length;
+      })
+      break;
 
-        case SortType.LENGTH:
-          return good1.length - good2.length;
-
-        default:
-          return 0;
-      }
-    });
+    case SortType.NONE:
+    default:
+      break;
   }
 
   if (isReversed) {
@@ -56,7 +58,7 @@ export function getReorderedGoods(
   }
 
   return visibleGoods;
-}
+};
 
 type State = {
   isReversed: boolean,


src/App.tsx
```diff
@@ -21,75 +22,147 @@ enum SortType {
...
+            )}
+            onClick={this.sortByLength}
+          >
+            Sort by length
+          </button>
+
+          <button
+            type="button"
+            className={classNames(
+              'button is-warning', { 'is-light': isReversed !== true },
```
DiDemkiv, line 138: can be simplified:
```suggestion
              'button is-warning', { 'is-light': !isReversed },
```

src/App.tsx
```diff
@@ -21,75 +22,147 @@ enum SortType {
...
+            )}
+            onClick={this.sortByAlphapet}
+          >
+            Sort alphabetically
+          </button>
+
+          <button
+            type="button"
+            className={classNames(
+              'button is-succes', { 'is-light': sortType !== SortType.LENGTH },
```
DiDemkiv, line 128: you misspelled `is-success` class name, therefore your ui doesnt work properly
```suggestion
              'button is-success', { 'is-light': sortType !== SortType.LENGTH },
```

DiDemkiv (changes requested):
small mistakes, but otherwise looks good to me 🔥

src/App.tsx
```diff
@@ -21,75 +22,147 @@ enum SortType {
...
+          return good1.length - good2.length;
+
+        default:
+          return 0;
+      }
+    });
+
+    if (isReversed) {
+      visibleGoods.reverse();
+    }
```
solaryasha, line 110: let's do it inside `getReorderedGoods` function

src/App.tsx
```diff
@@ -21,75 +22,147 @@ enum SortType {
...
+    const visibleGoods = [...goodsFromServer];
+
+    visibleGoods.sort((good1: string, good2: string) => {
+      switch (sortType) {
+        case SortType.ALPHABET:
+          return good1.localeCompare(good2);
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+
+        default:
```
solaryasha, line 103: It would be nice to use the switch first and then sort all the cases

solaryasha (changes requested):

src/App.tsx
```diff
@@ -34,62 +35,114 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
```
valerii-derkach, line 40: your reverse doesn't work correctly - it works only before sorting - check it
what is expected - I press sort by - then reverse is reversing sorted list

src/App.tsx
```diff
@@ -34,62 +35,114 @@ export function getReorderedGoods(
...
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      default:
+        return 0;
+    }
+  });
```
valerii-derkach, line 53: We don't need to perform sort if `sortType` is None.
So, it's better to move `switch` instruction above sort callback

valerii-derkach (changes requested):

src/App.tsx
```diff
@@ -34,62 +35,114 @@ export function getReorderedGoods(
...
+            ? (
+              <button
+                type="button"
+                className="button is-danger is-light"
+                onClick={this.reset}
+              >
+                Reset
+              </button>
+            )
+            : null}
```
valerii-derkach, line 135: use && here instead of ternary

valerii-derkach (changes requested):

ivan-kadykalo (changes requested):
Why you didn't fix Valerie's comments?

valerii-derkach (approved):

PR #1392, opened 2023-01-06
"list-of-goods" DiDemkiv

DiDemkiv:
[DEMO LINK](https://DiDemkiv.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


OlexandraKorol (approved):
As for me, looks good and clear)

src/App.tsx
```diff
@@ -26,70 +27,122 @@ type ReorderOptions = {
...
+  visibleGoods.sort((good1: string, good2: string): number => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
+
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      default:
+        return 0;
```
solaryasha, line 45: We don't need to perform sort if `sortType` is None.
So, it's better to move `switch` instruction above sort callback

solaryasha (approved):
👍

PR #1398, opened 2023-01-07
solution

Lun4trick:
[DEMO LINK](https://lun4trick.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -34,6 +35,25 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
+  if (sortType) {
+    visibleGoods.sort((good1, good2) => {
+      switch (sortType) {
+        case SortType.ALPHABET:
+          return good1.localeCompare(good2);
+
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+
+        default:
+          return 0;
+      }
+    });
+  }
+
+  if (isReversed) {
+    return visibleGoods.reverse();
+  }
+
   // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -34,6 +35,25 @@ export function getReorderedGoods(
...
+    });
+  }
+
+  if (isReversed) {
+    return visibleGoods.reverse();
+  }
+
   // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
```
ysrckr, line 59: Let's remove console statements in final code.

src/App.tsx
```diff
@@ -42,54 +62,103 @@ export function getReorderedGoods(
...
+      sortType,
+    } = this.state;
+
+    const {
+      NONE,
+      ALPHABET,
+      LENGTH,
+    } = SortType;
+
+    const orderedGoods = getReorderedGoods(goodsFromServer, this.state);
```
ysrckr, line 88: It would be better to have this function outside of the render.

src/App.tsx
```diff
@@ -42,54 +62,103 @@ export function getReorderedGoods(
...
+        <div className="buttons">
+          <button
+            type="button"
+            className={cn(
+              'button is-info',
+              { 'is-light': sortType !== ALPHABET },
+            )}
+            onClick={() => (
+              this.setState({ sortType: ALPHABET })
+            )}
```
ysrckr, line 101: It is better to create function for event. It is more readable that way.

src/App.tsx
```diff
@@ -42,54 +62,103 @@ export function getReorderedGoods(
...
+
+          <button
+            type="button"
+            className={cn(
+              'button is-success',
+              { 'is-light': sortType !== LENGTH },
+            )}
+            onClick={() => (
+              this.setState({ sortType: LENGTH })
+            )}
```
ysrckr, line 114: It is better to create function for event. It is more readable that way.

src/App.tsx
```diff
@@ -42,54 +62,103 @@ export function getReorderedGoods(
...
+
+          <button
+            type="button"
+            className={cn(
+              'button is-warning',
+              { 'is-light': !isReversed },
+            )}
+            onClick={() => (
+              this.setState({ isReversed: !isReversed })
+            )}
```
ysrckr, line 127: It is better to create function for event. It is more readable that way.

src/App.tsx
```diff
@@ -42,54 +62,103 @@ export function getReorderedGoods(
...
+            Reverse
+          </button>
+
+          {(sortType !== NONE || isReversed) && (
+            <button
+              type="button"
+              className="button is-danger is-light"
+              onClick={() => (
+                this.setState({ sortType: NONE, isReversed: false })
+              )}
```
ysrckr, line 138: It is better to create function for event. It is more readable that way.

ysrckr (changes requested):

src/App.tsx
dolnaro: I agree with that

dolnaro (approved):

src/App.tsx
```diff
@@ -42,54 +62,111 @@ export function getReorderedGoods(
...
+
+    const sortByLength = () => (
+      this.setState({ sortType: LENGTH })
+    );
+
+    const setIsReversed = () => (
+      this.setState({ isReversed: !isReversed })
+    );
+
+    const setDefault = () => (
```
naziamoff, line 100: I guess just `reset` would make more sense, but not bad as it is

src/App.tsx
```diff
@@ -42,54 +62,111 @@ export function getReorderedGoods(
...
+            className={cn(
+              'button is-warning',
+              { 'is-light': !isReversed },
+            )}
+            onClick={setIsReversed}
+          >
+            Reverse
+          </button>
+
+          {(sortType !== NONE || isReversed) && (
```
naziamoff, line 142: should be separated to a variable

src/App.tsx
```diff
@@ -42,54 +62,111 @@ export function getReorderedGoods(
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+
+            {
+            // eslint-disable-next-line max-len
+              orderedGoods.map(good => (
```
naziamoff, line 158: what's that? 😄

naziamoff, line 158:
```suggestion
            {orderedGoods.map(good => (
```

naziamoff (approved):

PR #1407, opened 2023-01-09
solution

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -15,11 +16,13 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
-}
+// enum SortType {
+//   NONE,
+//   ALPHABET,
+//   LENGTH,
+// }
+
+type SortType = 'none' | 'alphabet' | 'length' ;
 
 type ReorderOptions = {
   sortType: SortType,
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,11 +16,13 @@ export const goodsFromServer = [
...
-  ALPHABET,
-  LENGTH,
-}
+// enum SortType {
+//   NONE,
+//   ALPHABET,
+//   LENGTH,
+// }
+
+type SortType = 'none' | 'alphabet' | 'length' ;
```
Grayfox147, line 25: Interesting approach, you decided not to work with the enums, in case you wanted the enums you could reach them by using SortType.ALPHABET or SortType.LENGTH etc. but I dont believe it makes much difference, you can also create a type like you did.

Grayfox147 (commented):
Great Implementation, different approach in some things than mine but overall I believe it is an excellent job! :D

mmahfuzi:
> Great Implementation, different approach in some things than mine but overall I believe it is an excellent job! :D

Thanks :)

naziamoff (changes requested):

src/App.tsx
```diff
@@ -98,15 +95,16 @@ export class App extends React.Component<{}, State> {
 
     const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
 
+    const resetCheck = isReversed || sortType !== SortType.NONE;
```
naziamoff, line 98: > `resetCheck`

is this a function?

naziamoff (approved):
looks good 👍 
still you should pay attention to naming your variables

src/App.tsx
mmahfuzi: Hello @naziamoff, this is a variable that store multiple conditions for the rest, the first name that came to my mind :)

PR #1411, opened 2023-01-10
solution

kenvastolord:
[DEMO LINK](https://kenvastolord.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -33,17 +33,17 @@ export function getReorderedGoods(
 ) {
   const visibleGoods = [...goods];
 
-  switch (sortType) {
-    case SortType.ALPHABET:
-      visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
-      break;
-
-    case SortType.LENGTH:
-      visibleGoods.sort((good1, good2) => good1.length - good2.length);
-      break;
-
-    default:
-      break;
+  if (sortType) {
+    visibleGoods.sort((good1, good2) => {
+      switch (sortType) {
+        case SortType.ALPHABET:
+          return good1.localeCompare(good2);
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+        default:
+          return 0;
+      }
+    });
   }
 
   if (isReversed) {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
```
Grayfox147, line 38:
```suggestion
      if (sortType) {
    visibleGoods.sort((good1, good2=> {
      switch (sortType) {
        case SortType.ALPHABET:
          return good1.localeCompare(good2);

        case SortType.LENGTH:
          return good1.length - good2.length;

        default:
          return 0;
      }
    });
  }
```

src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+
+  render() {
+    const { sortType, isReversed } = this.state;
+
+    const visibleGoods = getReorderedGoods(
+      goodsFromServer,
+      { sortType, isReversed },
+    );
+
+    const isResetButtonVisible = (sortType !== SortType.NONE || isReversed);
```
Grayfox147, line 96: very good this is what we should do when we have a long checking comparison is better to store it like you did in a variable!

Grayfox147 (commented):
Overall! a great implementation! I suggested some changes but I believe is an excellent job!

src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+      isReversed: !prevState.isReversed,
+    }));
+  };
+
+  handleClickReset = () => {
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
```
naziamoff, line 86: You can keep it simple here 😄 
```suggestion
  sortAlphabetically = () => {
    this.setState({ sortType: SortType.ALPHABET });
  };

  sortByLength = () => {
    this.setState({ sortType: SortType.LENGTH });
  };

  reverse = () => {
    this.setState(prevState => ({
      isReversed: !prevState.isReversed,
    }));
  };

  reset = () => {
    this.setState({
      sortType: SortType.NONE,
      isReversed: false,
    });
  };
```

src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            onClick={this.handleClickSortAlphabetically}
+            className={classNames(
+              'button',
+              'is-info',
```
naziamoff, line 106:
```suggestion
              'button is-info',
```

src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+          >
+            Sort alphabetically
+          </button>
+
+          <button
+            type="button"
+            onClick={this.handleClickSortByLength}
+            className={classNames(
+              'button',
+              'is-success',
```
naziamoff, line 118: same here

naziamoff (approved):
well done! Just a couple of code style suggestions

PR #1417, opened 2023-01-10
Solution

Rijo-J:
- [DEMO LINK](https://rijo-j.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -53,6 +53,18 @@ export class App extends Component<{}, State> {
     sortType: SortType.NONE,
   };
 
+  handleSortByAlphab = () => {
+    this.setState(() => ({
+      sortType: SortType.ALPHABET,
+    }));
+  };
+
+  handleSortByLength = () => {
+    this.setState(() => ({
+      sortType: SortType.LENGTH,
+    }));
+  };
+
   handleReverse = () => {
     this.setState((state) => ({
       isReversed: !state.isReversed,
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +26,129 @@ type ReorderOptions = {
...
+      { sortType, isReversed },
+    );
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            onClick={this.handleSortByAlphab}
+            className={`button is-info ${
```
naziamoff, line 103: let's use [classnames](https://www.npmjs.com/package/classnames) library instead

src/App.tsx
```diff
@@ -26,70 +26,129 @@ type ReorderOptions = {
...
+            type="button"
+            onClick={this.handleReverse}
+            className={`button is-warning ${
+              !isReversed && 'is-light'
+            }`}
+          >
+            Reverse
+          </button>
+
+          {(isReversed || sortType !== SortType.NONE) && (
```
naziamoff, line 130: let's create a separate variable for that. You should do that for multiple conditions

naziamoff (changes requested):

src/App.tsx
Rijo-J: ✅

src/App.tsx
Rijo-J: ✅

PR #1418, opened 2023-01-10
list of all goods in the hoods

Kavster92:
[DEMO LINK](https://Kavster92.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -106,7 +106,7 @@ export class App extends React.PureComponent<{}, State> {
             )}
             onClick={this.sortByAlphabet}
           >
-            Sort Alphabetically
+            Sort alphabetically
           </Button>
 
           <Button
@@ -118,7 +118,7 @@ export class App extends React.PureComponent<{}, State> {
             )}
             onClick={this.sortByLength}
           >
-            Sort By Length
+            Sort by length
           </Button>
 
           <Button
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
+import classNames from 'classnames';
 import './App.scss';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -26,70 +29,121 @@ type ReorderOptions = {
...
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  if (sortType === SortType.ALPHABET) {
```
Grayfox147, line 38:
```suggestion
  if (sortType) {
    visibleGoods.sort((a, b) => {
      switch (sortType) {
        case SortType.ALPHABET:
          return a.localeCompare(b);

        case SortType.LENGTH:
          return a.length - b.length;

        default:
          return 0;
      }
    });
  }
```

Grayfox147 (commented):
Overall a great work, really smart implementation! I left a couple of suggestions!

src/App.tsx
naziamoff: sure, you always use switch if the number of conditions is limited

src/App.tsx
```diff
@@ -26,70 +29,121 @@ type ReorderOptions = {
...
+              'button is-info',
+              {
+                'is-light': !isReversed,
+              },
+            )}
+            onClick={this.reverse}
+          >
+            Reverse
+          </Button>
+          {(isReversed || sortType !== SortType.NONE) && (
```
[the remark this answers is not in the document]
naziamoff: agree with @Grayfox147 on this one

naziamoff (changes requested):

meforoma (approved):
✅

PR #1424, opened 2023-01-18
solution

nzrii-kysel:
[DEMO LINK](https://nzrii-kysel.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -50,6 +50,7 @@ export class App extends Component<{}, State> {
 
   render() {
     const { sortType, isReversed } = this.state;
+    const isResetButtonVisible = isReversed || sortType !== SortType.NONE;
 
     return (
       <div className="section content">
@@ -90,7 +91,7 @@ export class App extends Component<{}, State> {
             Reverse
           </button>
 
-          {(isReversed || sortType !== SortType.NONE)
+          {isResetButtonVisible
             && (
               <button
                 type="button"
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -90,7 +90,7 @@ export class App extends Component<{}, State> {
             Reverse
           </button>
 
-          {(this.state.isReversed || this.state.sortType !== SortType.NONE)
+          {(isReversed || sortType !== SortType.NONE)
             && (
               <button
                 type="button"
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -15,81 +20,94 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
-      </ul>
-    </div>
-  );
-};
+          {(this.state.isReversed || this.state.sortType !== SortType.NONE)
```
Malva37, line 93: `(isReversed || sortType !== SortType.NONE)`
 it will also work

you already did destructurisation:
`const { sortType, isReversed } = this.state;`

Malva37 (approved):
great

src/App.tsx
```diff
@@ -15,81 +20,94 @@ export const goodsFromServer = [
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
-      </ul>
-    </div>
-  );
-};
+          {(isReversed || sortType !== SortType.NONE)
```
Anna-Tkachenko, line 93: will be better to create separate boolean variable before the return statement with some understandable name
instead of combining several checks here and according to them show some component

src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,42 @@
...
+      renderedGoods.sort((a, b) => a.length - b.length);
+      break;
+
+    default:
+      // do nothing
+  }
+
+  if (isReversed) {
+    renderedGoods.reverse();
+  }
```
Anna-Tkachenko, line 28: not important, but will be great to create separate function for list transforming, that will accept needed parameters and will return the refactored list

Anna-Tkachenko (approved):

Anna-Tkachenko (changes requested):
there was initial `getReorderedGoods` function, that you need to use for transforming the list, rewrite your code using those function

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #1435, opened 2023-01-26
solution

AnnyKoval:
[DEMO LINK](https://AnnyKoval.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,119 @@ type ReorderOptions = {
...
+    }));
+  };
+
+  reset = () => {
+    this.setState({ sortType: SortType.NONE, isReversed: false });
+  };
+
+  render() {
+    const { sortType, isReversed } = this.state;
+
```
Bohdan-Maksymiuk, line 87: better for readability get reorderedGoods here in variable and then map in return

Bohdan-Maksymiuk (approved):
Good job 👍

PR #1436, opened 2023-01-26
React list of goods - completed

GarryRocksman:
[DEMO](https://GarryRockaman.github.io/react_list-of-goods/)

 - Have a nice Day 🔥

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


Avichez (approved):
That's a great solution 🔥 
Especially nice idea with `Math.random` in this situation, but according to react documentation better to avoid this method.
here is some info - https://reactjs.org/docs/reconciliation.html#tradeoffs

src/App.tsx
```diff
@@ -32,64 +33,136 @@ export function getReorderedGoods(
...
+  };
+
+  reset = () => {
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    });
+  };
+
+  sortAlphabet = () => {
```
ArturErmolenko, line 88:
```suggestion
  sortByAlphabet = () => {
```

src/App.tsx
```diff
@@ -32,64 +33,136 @@ export function getReorderedGoods(
...
+            className={classNames(
+              'button is-warning',
+              { 'is-light': !isReversed },
+            )}
+            onClick={this.reverse}
+          >
+            Reverse
+          </button>
+
+          {(sortType !== SortType.NONE || isReversed)
```
ArturErmolenko, line 143: store in variable and use it here

ArturErmolenko (approved):
GJ 👍

PR #1441, opened 2023-01-28
solution

Avichez:
[DEMO LINK](https://Avichez.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
@@ -34,6 +35,21 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
+  switch (sortType) {
+    case 1:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+    case 2:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+    default:
+      break;
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -34,6 +35,21 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
```
ArturErmolenko, line 43: DRY, use sort just once and use switch case in callback

src/App.tsx
```diff
@@ -42,54 +58,95 @@ export function getReorderedGoods(
...
+            className={classNames(
+              'button is-warning',
+              { 'is-light': !isReversed },
+            )}
+            onClick={this.reverseListOfGoods}
+          >
+            Reverse
+          </button>
+
+          {(sortType !== SortType.NONE || isReversed) && (
```
ArturErmolenko, line 133: store it to some  variable & use here

src/App.tsx
```diff
@@ -34,6 +35,21 @@ export function getReorderedGoods(
...
   const visibleGoods = [...goods];
 
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+    default:
```
ArturErmolenko, line 45: throw error in default

src/App.tsx
```diff
@@ -42,54 +58,95 @@ export function getReorderedGoods(
...
+  resetListOfGoods = () => {
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
```
ArturErmolenko, line 95: 👍

ArturErmolenko (changes requested):

src/App.tsx
Avichez: @ArturErmolenko  Is that good to use `switch case` many times inside of an iterator instead of one time and then start the iterator?

ArturErmolenko: you use switch just once & passing right callback into sort method

ArturErmolenko (approved):
Well done 👍

src/App.tsx
Avichez: Oh, got it, thank you 🙂

PR #1452, opened 2023-01-31
solution

JOWISSA:
https://JOWISSA.github.io/react_list-of-goods/


src/App.tsx
```diff
@@ -26,70 +27,143 @@ type ReorderOptions = {
...
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
+  if (SortType) {
```
tetiana-ivanchenko, line 36: what is the point in this condition?

tetiana-ivanchenko (changes requested):

src/App.tsx
nazarmatsevych: This condition looks redundant, so you should remove it, but you didnt add any changes. If you have any questions about comments - ask them in fe_chat and try to solve all of the problems before making re-request.

nazarmatsevych: *

nazarmatsevych (changes requested):

src/App.tsx
JOWISSA: то я забувся зробити коміт))

> This condition looks redundant, so you should remove it, but you didnt add any changes. If you have any questions about comments - ask them in fe_chat and try to solve all of the problems before making re-request.

vlad-demchuk (approved):

PR #1457, opened 2023-02-01
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,120 @@ type ReorderOptions = {
...
+        }
+      });
+
+      if (isReversed) {
+        visibleGoods.reverse();
+      }
+
+      // eslint-disable-next-line no-console
+      console.log(sortType, isReversed);
+
```
denys-cheporniuk, line 86: remove logs from final code

src/App.tsx
```diff
@@ -26,70 +27,120 @@ type ReorderOptions = {
...
+          case SortType.LENGTH:
+            return gA.length - gB.length;
+          default:
+            return 0;
+        }
+      });
+
+      if (isReversed) {
+        visibleGoods.reverse();
+      }
```
denys-cheporniuk, line 82: this function shouldn't be here

src/App.tsx
```diff
@@ -26,70 +27,120 @@ type ReorderOptions = {
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {getReorderedGoods(goodsFromServer, this.state).map(good => {
+              return <li data-cy="Good" key={good}>{good}</li>;
+            })}
```
denys-cheporniuk, line 140: call getReorderedGoods function between render and return instead of calling it here

denys-cheporniuk (changes requested):

src/App.tsx
```diff
@@ -26,70 +27,119 @@ type ReorderOptions = {
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {reorderedGoods.map(good => {
+              return <li data-cy="Good" key={good}>{good}</li>;
```
Anna-Tkachenko, line 138: `return` statement is redundant here

Anna-Tkachenko (approved):

PR #1458, opened 2023-02-01
List was created

Coyoote:
[DEMO LINK](https://Coyoote.github.io/react_list-of-goods/)

Have a nice day!

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+    case SortType.ALPHABET:
+      visibleGoods.sort();
+      break;
+
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+
+    default:
+  }
```
Anastasiia-Khudych, line 46: It looks like you forgot to write the default case.

Anastasiia-Khudych (approved):
Good job!

src/App.tsx
Coyoote: I tried to add something but i got a diffirent error each time

Anastasiia-Khudych: Have you tried `return 0` ?

Anastasiia-Khudych: I think it's better to write `switch` inside the `sort` method, you will not repeat code with sort several times and you can make a normal default case like `return 0`.

Coyoote: Return 0 returns an error, switch inside sort sound good

ArturErmolenko (approved):

PR #1462, opened 2023-02-01
solution to list-of-goods task

andrii-kurash:
[DEMO LINK](https://andrii-kurash.github.io/react_list-of-goods/)

Good luck!✨

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -31,65 +32,123 @@ export function getReorderedGoods(
...
-      <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {getReorderedGoods(goodsFromServer, this.state).map(good => (
```
denys-cheporniuk, line 141: it's better to call getReorderedGoods function between render and return instead of calling it here

denys-cheporniuk (approved):

PR #1466, opened 2023-02-01
Solution

maasiimka:


src/App.tsx
```diff
@@ -34,10 +35,24 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
+  visibleGoods.sort((a, b) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return a < b ? -1 : 1;
```
Anastasiia-Khudych, line 41: I think it's better to use `localeCompare` for strings comparing.

src/App.tsx
```diff
@@ -47,49 +62,110 @@ export function getReorderedGoods(
...
+  };
+
+  handleSortType = (order: SortType) => {
+    this.setState({ sortType: order });
+  };
+
+  handleOrderChange = () => {
+    this.setState((state) => ({
+      ...state,
+      isReversed: !state.isReversed,
```
Anastasiia-Khudych, line 78: `State` destructuring is useless here

src/App.tsx
Anna-Tkachenko: agree, fix @Anastasiia-Khudych comment

src/App.tsx
Anna-Tkachenko: fix @Anastasiia-Khudych comment
the process of copying all properties from state is processed inside `setState` function

src/App.tsx
```diff
@@ -47,49 +62,110 @@ export function getReorderedGoods(
...
+              'is-warning ',
+              { 'is-light': !isReversed },
+            )}
+            onClick={handleOrderChange}
+          >
+            Reverse
+          </button>
+
+          {
+            (sortType !== SortType.NONE || isReversed) && (
```
Anna-Tkachenko, line 149: will be better to create separate boolean variable before the `return` statement with the result of this check instead of combining several checks directly in `jsx`
in this way you code will be clearer

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1473, opened 2023-02-01
solution

BrandonSunrise:
[DEMO LINK](https://BrandonSunrise.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,130 @@ type ReorderOptions = {
...
+          <button
+            type="button"
+            className={classNames('button is-warning', {
+              'is-light': !isReversed,
+            })}
+            onClick={this.reverseGoods}
+          >
+            Reverse
+          </button>
+          {(isReversed || sortType !== SortType.NONE)
```
anna-inozemets, line 136: You can create a bollean variable for this statement, maybe it will be useful when you will have one more check like that (in the future)

anna-inozemets (approved):
Really good solution!🐇

src/App.tsx
BrandonSunrise: Thank you for your coment! I've already fix it!

ArturErmolenko (approved):
Well done 👍

PR #1476, opened 2023-02-01
solution

Dashaverz:
[DEMO LINK](https://Dashaverz.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -35,6 +36,24 @@ export function getReorderedGoods(
   const visibleGoods = [...goods];
 
   // Sort and reverse goods if needed
+
+  visibleGoods.sort((a, b) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return a.localeCompare(b);
+
+      case SortType.LENGTH:
+        return a.length - b.length;
+
+      default:
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    return visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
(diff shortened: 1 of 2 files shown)


Anton-Zhytonbaiev (approved):
GJ 🔥

src/App.tsx
```diff
@@ -35,6 +36,24 @@ export function getReorderedGoods(
...
+  visibleGoods.sort((a, b) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return a.localeCompare(b);
+
+      case SortType.LENGTH:
+        return a.length - b.length;
+
+      default:
+        return 0;
```
ArturErmolenko, line 49: throw error in default case

src/App.tsx
```diff
@@ -47,49 +66,92 @@ export function getReorderedGoods(
...
+            type="button"
+            className={classNames('button is-warning', {
+              'is-light': !isReversed,
+            })}
+            onClick={this.reverse}
+          >
+            Reverse
+          </button>
+
+          {(sortType || isReversed) && (
```
ArturErmolenko, line 136: store this expression in variable and result here

ArturErmolenko (changes requested):

src/App.tsx
```diff
@@ -35,6 +36,24 @@ export function getReorderedGoods(
   const visibleGoods = [...goods];
 
   // Sort and reverse goods if needed
+
+  visibleGoods.sort((a, b) => {
+    switch (sortType) {
```
ArturErmolenko, line 41: what happened if we pass SortType.NONE?

ArturErmolenko (changes requested):

src/App.tsx
```diff
@@ -35,6 +36,25 @@ export function getReorderedGoods(
...
+      case SortType.ALPHABET:
+        return a.localeCompare(b);
+
+      case SortType.LENGTH:
+        return a.length - b.length;
+
+      case SortType.NONE:
+      default:
+        throw new Error('Unexpected sortBy value');
```
ArturErmolenko, line 50: SortType.None is expected value, in this case you just shouldn't sort array, so you could return 0 , error should throw if we pass value that doesn't exist in enum

ArturErmolenko (changes requested):

ArturErmolenko (approved):
GJ 👍

PR #1479, opened 2023-02-02
solution

OlehSharha:
[DEMO LINK](https://OlehSharha.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -34,62 +35,130 @@ export function getReorderedGoods(
...
 
+  visibleGoods.sort((good1, good2) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+      case SortType.NONE:
+      default:
+        return 0;
```
vanst1, line 49:
```suggestion
      case SortType.NONE:
        return 0;
        
        default:
        `throw error here`
```

vanst1 (changes requested):
Good job) 
Some fix is needed:)

Fix color of the buttons
Screenshot_1

src/App.tsx
```diff
@@ -34,62 +35,131 @@ export function getReorderedGoods(
...
+    this.setState({
+      sortType: SortType.LENGTH,
+    })
+  );
+
+  sortByAlphabet = () => (
+    this.setState({
+      sortType: SortType.ALPHABET,
+    })
+  );
```
Anna-Tkachenko, line 83: these functions shouldn't return anything, check correct braces
```suggestion
  sortByAlphabet = () => {
    this.setState({
      sortType: SortType.ALPHABET,
    });
  };
```

Anna-Tkachenko (approved):

src/App.tsx
OlehSharha: Thx for comment,  i fixed braces and commited changes)

PR #1482, opened 2023-02-02
solution

metalikaqq:
[DEMO LINK](https://metalikaqq.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -34,62 +35,126 @@ export function getReorderedGoods(
...
-        </button>
-      </div>
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    isReversed: true,
```
tania-boiar, line 69: I think it's better to use `false` as the default for `isReversed`.

src/App.tsx
```diff
@@ -34,62 +35,126 @@ export function getReorderedGoods(
...
+          >
+            Reverse
+          </button>
+
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={this.handleReset}
+          >
+            Reset
```
tania-boiar, line 144: You have to show this button only when the other one (sort or reverse) is selected.

tania-boiar (changes requested):
Good job!)

You did not pass all the tests.
I left you some comments that I think could be the reasons for that.

PR #1492, opened 2023-02-06
Initial commit

yaroslav-furhala:
- [DEMO LINK](https://yaroslav-furhala.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,7 @@
+import classNames from 'classnames';
 import React from 'react';
-import 'bulma/css/bulma.css';
-import './App.scss';
 
-export const goodsFromServer = [
+const goodsFromServer: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,81 +14,136 @@ export const goodsFromServer = [
...
+      isReversed: !state.isReversed,
+    }));
+  };
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  sortAlphabetically = () => {
+    this.setState({
+      sortBy: 'alpha',
```
loralevitska, line 37: What is alpha? Use more meaningful variable names

src/App.tsx
```diff
@@ -15,81 +14,136 @@ export const goodsFromServer = [
...
+      visibleGoods.sort((a, b) => {
+        switch (sortBy) {
+          case 'alpha':
+            return a.localeCompare(b);
+          case 'length':
+            return a.length - b.length;
+          default:
+            return 0;
+        }
+      });
```
loralevitska, line 74:
```suggestion
        switch (sortBy) {
          case 'alpha':
            return a.localeCompare(b);
            
          case 'length':
            return a.length - b.length;
            
          default:
            return 0;
        }
      });
```

src/App.tsx
```diff
@@ -15,81 +14,136 @@ export const goodsFromServer = [
...
+              Start
+            </button>
+          )
+          : (
+            <>
+              <ul className="goods-list">
+                {visibleGoods.map(good => (
+                  <li key={good} className="goods-list__item">{good}</li>
+                ))}
+              </ul>
```
loralevitska, line 103: Divide your solution down into separate components

src/App.tsx
```diff
@@ -15,81 +14,136 @@ export const goodsFromServer = [
...
+              <div className="buttons">
```
loralevitska, line 104: same here `separate component`

src/App.tsx
```diff
@@ -15,81 +14,136 @@ export const goodsFromServer = [
...
+      >
+        {!start
+          ? (
+            <button
+              type="button"
+              onClick={this.show}
+              className="button button--start"
+            >
+              Start
+            </button>
```
loralevitska, line 95: For what this button? Maybe that's why the test doesn't pass, because it doesn't see the data at first?

loralevitska (changes requested):
Good job! Let's improve
Try to pass all tests 
image

yaroslav-furhala:
1. (Try to pass all tests) - I can`t pass all test because my cypress broken ( I reinstaled node modules, and this doesn`t help me to fix problem. I`m ask on Q&A about this problem and mentor said, it`s can`t posible to fix, because it is cypress bug!)
2. (What is alpha? Use more meaningful variable names) - Fixed!
3. (For what this button? Maybe that's why the test doesn't pass, because it doesn't see the data at first?) - I don`t think so, because a have problems with CYPRESS !

yaroslav-furhala:
And, my attribute data-cy="Good" is on my code, but CYPRESS have bug !

SanyaBratashchuk (approved):

PR #1495, opened 2023-02-07
Solution

offsset:
https://offsset.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,117 @@ type ReorderOptions = {
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {visibleGoods1.map(good => (
+              <li data-cy="Good">{good}</li>
```
tetiana-ivanchenko, line 136: list items should have key attr

src/App.tsx
```diff
@@ -26,70 +27,117 @@ type ReorderOptions = {
...
+};
+
+export class App extends React.Component <{}, State> {
+  state: Readonly<State> = {
+    sortType: SortType.NONE,
+    isReversed: false,
+  };
+
+  render() {
+    const visibleGoods1 = getReorderedGoods(goodsFromServer, this.state);
```
tetiana-ivanchenko, line 66: use another naming instead of adding number to the name

tetiana-ivanchenko (changes requested):

src/App.tsx
```diff
@@ -26,70 +27,116 @@ type ReorderOptions = {
...
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((a, b) => {
+    switch (sortType) {
+      case 1:
+        return a.localeCompare(b);
+      case 2:
```
nazarmatsevych, line 40: It's better to use values from ENUM instead of case 1 and case 2

nazarmatsevych (changes requested):

nazarmatsevych (approved):

PR #1504, opened 2023-02-15
solution

Deniskakaka:
https://deniskakaka.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,70 +26,105 @@ type ReorderOptions = {
   isReversed: boolean,
 };
 
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
+  if (sortType === SortType.ALPHABET) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
-  return visibleGoods;
+  return isReversed ? visibleGoods.reverse() : visibleGoods;
 }
 
-// DON'T save goods to the state
-// type State = {
-//   isReversed: boolean,
-//   sortType: SortType,
-// };
-
-export const App: React.FC = () => {
-  return (
-    <div className="section content">
-      <div className="buttons">
-        <button
-          type="button"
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,70 +26,105 @@ type ReorderOptions = {
...
+  };
+
+  render() {
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={`button is-info ${this.state.sortType !== SortType.ALPHABET && 'is-light'}`}
+            onClick={() => this.setState((prev: State) => {
```
VitaliyBondarenko1982, line 67: Bad practice to place anonimous arrow functions in JSX.
Please make funcitons and pass here and in other places reference on functions.

src/App.tsx
```diff
@@ -26,70 +26,105 @@ type ReorderOptions = {
...
+                sortType: SortType.ALPHABET,
+              };
+            })}
+          >
+            Sort alphabetically
+          </button>
+
+          <button
+            type="button"
+            className={`button is-success ${this.state.sortType !== SortType.LENGTH && 'is-light'}`}
```
VitaliyBondarenko1982, line 79: https://www.npmjs.com/package/classnames
convinient library for making conditional classnames
import cn from 'classnames'

```suggestion
            className={cn(`button  is-success', { 'is-light': this.state.sortType !== SortType.LENGTH }}
```

VitaliyBondarenko1982: Will be something like that, check other places

VitaliyBondarenko1982 (changes requested):

src/App.tsx
```diff
@@ -26,70 +27,133 @@ type ReorderOptions = {
...
+  sortType: SortType,
+};
+
+export class App extends React.PureComponent<{}, State> {
+  state: Readonly<State> = {
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
+
+  changeSatate(typeMove?: string, typeSort?: string) {
```
IvanFesenko, line 61: to avoid anonymous fuction use you can do this way
```suggestion
  changeState = (typeMove?: string, typeSort?: string) => () => {
```

src/App.tsx
```diff
@@ -26,70 +27,133 @@ type ReorderOptions = {
...
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={
+              classNames('button is-info', {
+                'is-light': this.state.sortType !== SortType.ALPHABET,
+              })
+            }
+            onClick={() => this.changeSatate('sort', 'alpha')}
```
IvanFesenko, line 109: And usage this way
```suggestion
            onClick={this.changeSatate('sort', 'alpha')}
```

IvanFesenko (approved):

PR #1508, opened 2023-02-21
Solves

Vasyliy22:
[DEMO LINK](https://Vasyliy22.github.io/react_list-of-goods/)
I didn`t use function getReorderedGoods because I didn`t understand how use in the rehder, Please explain me it

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,125 @@ type ReorderOptions = {
...
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {goods.map(list => (
+              <li data-cy="Good" key={list}>
+                {list}
+              </li>
+            ))}
+          </ul>
```
etojeDenys, line 146:
```suggestion
            {goods.map(list => (
              <li data-cy="Good" key={list}>
                {list}
              </li>
            ))}
```

src/App.tsx
```diff
@@ -26,70 +27,125 @@ type ReorderOptions = {
...
+      goods.sort();
+    }
+
+    if (sortType === SortType.LENGTH) {
+      goods.sort((a, b) => a.length - b.length);
+    }
+
+    if (isReversed) {
+      goods.reverse();
+    }
```
etojeDenys, line 86: move it into `getReorderedGoods`

etojeDenys (changes requested):

Vasyliy22:
Thanks👍👍

src/App.tsx
```diff
@@ -26,70 +27,120 @@ type ReorderOptions = {
...
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
+  if (sortType === SortType.ALPHABET) {
+    goods.sort();
+  }
```
polosanya, line 38: Sort is a mutable method, so it's better to sort a copy of goods(visibleGoods).
Also, use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) method inside the sort callback to correctly sort goods alphabetically.

src/App.tsx
```diff
@@ -26,70 +27,120 @@ type ReorderOptions = {
...
+
+  if (sortType === SortType.LENGTH) {
+    goods.sort((a, b) => a.length - b.length);
+  }
+
+  if (isReversed) {
+    goods.reverse();
```
polosanya, line 45: Reverse is a mutable method too

src/App.tsx
```diff
@@ -26,70 +27,120 @@ type ReorderOptions = {
...
+    this.setState({ sortType: SortType.ALPHABET });
+  };
+
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    const goods = [...goodsFromServer];
```
polosanya, line 85: No need to make a copy here too. You do it inside getReorderedGoods function. 
Also, save the result of that function to some variable

src/App.tsx
```diff
@@ -26,70 +27,120 @@ type ReorderOptions = {
...
+                <button
+                  type="button"
+                  className="button is-danger is-light"
+                  onClick={this.reset}
+                >
+                  Reset
+                </button>
+              )
+              : ''
+          }
```
polosanya, line 134:
```suggestion
          {sortType || isReversed
              && (
                <button
                  type="button"
                  className="button is-danger is-light"
                  onClick={this.reset}
                >
                  Reset
                </button>
          )}
```
You can also do it like this. Also, I think it's better to check if sortType !== SortType.None

polosanya (changes requested):
Let's improve your code a bit)

anstsot (approved):
Good job, approved!

PR #1510, opened 2023-02-23
React_list-of-goods

07Spirit:
[DEMO LINK](https://07Spirit.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,95 +1,116 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+import goodsFromServer from './server/goodsFromServer.json';
+import { GoodList } from './components/GoodList';
 
-export const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
-];
-
-enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
-}
-
-type ReorderOptions = {
-  sortType: SortType,
+type State = {
+  sortType: string,
   isReversed: boolean,
 };
 
-// Use this function in the render method to prepare goods
-export function getReorderedGoods(
-  goods: string[],
-  { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
+export class App extends React.Component<{}, State> {
+  state = {
+    sortType: 'NONE',
+    isReversed: false,
+  };
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  render() {
+    const { sortType, isReversed } = this.state;
+    const visibleGoods = [...goodsFromServer];
 
-  return visibleGoods;
-}
+    visibleGoods.sort((g1, g2) => {
+      switch (sortType) {
+        case 'ALPHABET':
+          return g1.localeCompare(g2);
 
-// DON'T save goods to the state
-// type State = {
-//   isReversed: boolean,
-//   sortType: SortType,
-// };
+        case 'LENGTH':
+          return g1.length - g2.length;
 
-export const App: React.FC = () => {
-  return (
-    <div className="section content">
-      <div className="buttons">
-        <button
-          type="button"
-          className="button is-info is-light"
-        >
-          Sort alphabetically
-        </button>
+        case 'NONE':
+        default:
+          return 0;
+      }
+    });
 
-        <button
-          type="button"
-          className="button is-success is-light"
-        >
-          Sort by length
-        </button>
+    if (isReversed) {
+      visibleGoods.reverse();
+    }
 
-        <button
-          type="button"
-          className="button is-warning is-light"
-        >
-          Reverse
-        </button>
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={
+              classNames(
+                'button',
+                'is-info',
+                { 'is-light': this.state.sortType !== 'ALPHABET' },
+              )
+            }
+            onClick={() => this.setState({ sortType: 'ALPHABET' })}
+          >
+            Sort alphabetically
+          </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+          <button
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,95 +1,116 @@
...
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  render() {
+    const { sortType, isReversed } = this.state;
+    const visibleGoods = [...goodsFromServer];
 
-  return visibleGoods;
-}
+    visibleGoods.sort((g1, g2) => {
```
agniya-i, line 23: it would be better to create a separate function `getReorderedGoods` outside the class for this

agniya-i (changes requested):

07Spirit:
Remake code for good result of test.

07Spirit:
But earler version work too, and have less code!

src/App.tsx
```diff
@@ -1,95 +1,129 @@
...
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((g1, g2) => {
+    switch (sortType) {
+      case 'ALPHABET':
```
polosanya, line 27: It would be great to create Enum with all possible sortTypes, so you don't need to hardcore them

src/App.tsx
```diff
@@ -1,95 +1,129 @@
...
-        </button>
+          <button
+            type="button"
+            className={classNames('button', 'is-success', {
+              'is-light': this.state.sortType !== 'LENGTH',
+            })}
+            onClick={() => this.setState({
+              sortType: 'LENGTH',
+              isResetActive: true,
+            })}
```
polosanya, line 78: Also, it would be great to create separate functions to handle these clicks

src/App.tsx
```diff
@@ -1,95 +1,129 @@
...
+                isReversed: !state.isReversed,
+                isResetActive: !state.isResetActive,
+              };
+            })}
+          >
+            Reverse
+          </button>
+
+          {this.state.isResetActive
+          && (
```
polosanya, line 106:
```suggestion
          {this.state.isResetActive && (
```
You could also destructure isResetActive from state

src/components/GoodList.tsx
```diff
@@ -0,0 +1,11 @@
+export const GoodList: React.FC<{ goods: string[] }> = ({ goods }) => (
```
polosanya, line 1: Why do you use functional component here and the class component in App?
Use only one thing

src/components/GoodList.tsx
```diff
@@ -0,0 +1,11 @@
...
+  <>
+    <ul>
+      {
+        goods.map((good) => (
+          <li data-cy="Good" key={good}>{good}</li>
+        ))
+      }
+    </ul>
+  </>
```
polosanya, line 10:
```suggestion
    <ul>
      {goods.map((good) => (
          <li data-cy="Good" key={good}>{good}</li>
      ))}
    </ul>
```
You don't need react fragment here, as you have ul wrap

polosanya (changes requested):
Let's improve your code a bit)

src/App.tsx
```diff
@@ -1,95 +1,148 @@
...
+          sortType: type,
+        });
+        break;
+
+      default:
+        break;
+    }
+  };
+
+  eventReverse = () => {
```
Volodymyr-Melnyk-Fron10der, line 77:
```suggestion
  handleReverseEvent = () => {
```

src/App.tsx
```diff
@@ -1,95 +1,148 @@
...
+    this.setState((state) => {
+      if (state.sortType !== SortType.NONE) {
+        return {
+          isReversed: !state.isReversed,
+        };
+      }
+
+      return {
+        isReversed: !state.isReversed,
+      };
```
Volodymyr-Melnyk-Fron10der, line 87: If I clearly see - if block returns the same result as bellow return, no?) 
```suggestion
      return {
        isReversed: !state.isReversed,
      };
```

src/App.tsx
```diff
@@ -1,95 +1,148 @@
...
+    });
+  };
+
+  render() {
+    const isReset
+    = this.state.sortType !== SortType.NONE || this.state.isReversed;
+
```
Volodymyr-Melnyk-Fron10der, line 94:
```suggestion
  render() {
    const { sortType, isReversed } = this.state;
    const isReset = sortType !== SortType.NONE || isReversed;

```

src/App.tsx
```diff
@@ -1,95 +1,148 @@
...
-          Reset
-        </button>
-      </div>
+export class App extends React.Component<{}, State> {
+  state = {
+    sortType: SortType.NONE,
+    isReversed: false,
+  };
+
+  eventSort = (type: SortType) => {
```
Volodymyr-Melnyk-Fron10der, line 56: if function will be passed in prop named like `on... (onClick, onChanged, onSelect, ...)` - the function should be named such a `handle...`  If it's just a helper function and it doesn't belong to the JSX- you can name it without `handle `prefix
```suggestion
  handleSortEvent = (type: SortType) => {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Good job, I left some suggestions

src/App.tsx
07Spirit: Thanks

src/App.tsx
```diff
@@ -1,95 +1,142 @@
...
+        break;
+    }
+  };
+
+  handleReverseEvent = () => {
+    this.setState((state) => {
+      return {
+        isReversed: !state.isReversed,
+      };
+    });
```
Anastasiia-Svintsova, line 82: Could be shorten
```suggestion
    this.setState((state) => ({
      isReversed: !state.isReversed,
    });
```

src/App.tsx
```diff
@@ -1,95 +1,142 @@
...
+            className={classNames('button', 'is-warning', {
+              'is-light': !this.state.isReversed,
+            })}
+            onClick={this.handleReverseEvent}
+          >
+            Reverse
+          </button>
+
+          {isReset
+          && (
```
Anastasiia-Svintsova, line 123:
```suggestion
          {isReset && (
```

Anastasiia-Svintsova (approved):

src/App.tsx
07Spirit: Thanks)

PR #1511, opened 2023-02-23
develop

barantarasNEW:
https://barantarasNEW.github.io/react_list-of-goods/


OlEzhA0 (approved):
hey! I approve your work but pls do next tasks with functional components, if no requirements to use classes. Would be better to rewrite even this task with FC

PR #1514, opened 2023-02-25
Solution by DK

DimaK123:
[DEMO LINK](https://DimaK123.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -34,62 +35,133 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
+  if (sortType === SortType.ALPHABET) {
+    visibleGoods.sort();
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
```
polosanya, line 43: You have a limited amount of sortTypes, so you can use switch...case here instead.

src/App.tsx
```diff
@@ -34,62 +35,133 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
+  if (sortType === SortType.ALPHABET) {
+    visibleGoods.sort();
```
polosanya, line 39: It is better to use localeCompare method to sort goods correctly alphabetically

polosanya (changes requested):
Almost done! Let's improve your code a bit)

src/App.tsx
DimaK123: Done

src/App.tsx
DimaK123: Done

src/App.tsx
```diff
@@ -34,62 +35,137 @@ export function getReorderedGoods(
...
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+
+    default:
+  }
```
polosanya, line 48: Add at least "break" or throw some error on default case

src/App.tsx
```diff
@@ -34,62 +35,137 @@ export function getReorderedGoods(
...
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
+  if (sortType === SortType.NONE && isReversed !== true) {
+    return goodsFromServer;
+  }
```
polosanya, line 58: I guess it's better to check it before switch...case

polosanya (approved):
Good job!

PR #1519, opened 2023-03-01
solution

bohdan-gavrilyk:
https://bohdan-gavrilyk.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,133 @@ type ReorderOptions = {
...
+    visibleGoods.sort((g1, g2) => {
+      return g1.localeCompare(g2);
+    });
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((g1, g2) => {
+      return g1.length - g2.length;
+    });
+  }
```
vlad-demchuk, line 46: let's use switch case

vlad-demchuk (approved):

PR #1526, opened 2023-03-04
solution

webdevw:
[DEMO LINK](https://webdevw.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -16,9 +17,9 @@ export const goodsFromServer = [
 ];
 
 enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
+  NONE = 'n',
+  ALPHABET = 'a',
+  LENGTH = 'l',
 }
 
 type ReorderOptions = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -16,80 +17,171 @@ export const goodsFromServer = [
...
+  length: boolean,
+};
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    isReversed: false,
+    sortType: SortType.NONE,
+    reset: false,
+    alphabet: false,
+    length: false,
```
agniya-i, line 70: do we need those extra fields in the state, if we have this data in sortType?

src/App.tsx
```diff
@@ -16,80 +17,171 @@ export const goodsFromServer = [
...
+            }
+            onClick={() => {
+              this.reverse();
+            }}
+          >
+            Reverse
+          </button>
+
+          {
+            (reset || isReversed)
```
agniya-i, line 159:
```suggestion
            (sortType !== SortType.NONE || isReversed)
```

here we can also simply check the state of the sort type

src/App.tsx
```diff
@@ -16,80 +17,171 @@ export const goodsFromServer = [
 ];
 
 enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
+  NONE = 'n',
+  ALPHABET = 'a',
+  LENGTH = 'l',
```
agniya-i, line 22:
```suggestion
  NONE = 'none',
  ALPHABET = 'alphabet',
  LENGTH = 'length',
```
better not to use short names

agniya-i (changes requested):

src/App.tsx
```diff
@@ -16,80 +17,159 @@ export const goodsFromServer = [
...
+    switch (sortType) {
+      case 'none':
+        return 0;
+      case 'alphabet':
+        return a.localeCompare(b);
+      case 'length':
+        return a.length - b.length;
+      default:
+        return 0;
+    }
```
anstsot, line 46: you can combine none and default cases
and it would be better to use power of TS to check the types

```suggestion
    switch (sortType) {
      case SortType.ALPHABET:
        return a.localeCompare(b);
      case SortType.LENGTH:
        return a.length - b.length;
      case SortType.NONE:
      default:
        return 0;
    }
```

anstsot (approved):
Great job, approved
You can add some improvement (check comments)

PR #1532, opened 2023-03-07
Solution

Vasyl-Pavlenko:
[DEMO LINK](https://vasyl-pavlenko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -33,17 +33,13 @@ export function getReorderedGoods(
 ) {
   const visibleGoods = [...goods];
 
-  if (sortType) {
+  if (sortType !== SortType.NONE) {
     visibleGoods.sort((good1, good2) => {
       if (sortType === SortType.LENGTH) {
         return good1.length - good2.length;
       }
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,70 +27,138 @@ type ReorderOptions = {
...
+        return good1.length - good2.length;
+      }
+
+      if (sortType === SortType.ALPHABET) {
+        return good1.localeCompare(good2);
+      }
+
+      return 0;
+    });
+  }
```
IvanFesenko, line 48: No need to sort if sortType is NONE
```suggestion
  if (sortType !== SortType.NONE) {
    visibleGoods.sort((good1, good2) => {
      if (sortType === SortType.LENGTH) {
        return good1.length - good2.length;
      }

      return good1.localeCompare(good2);
    });
  }
```

IvanFesenko (changes requested):
let add one improvement

IvanFesenko (approved):

PR #1533, opened 2023-03-07
result

novak-k:
[DEMO LINK](https://novak-k.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,118 @@ type ReorderOptions = {
...
-  console.log(sortType, isReversed);
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+    default:
+    // do nothing
```
tetiana-ivanchenko, line 44: remove comment and add error handling

tetiana-ivanchenko (changes requested):

src/App.tsx
novak-k: it is not an error, it is just sorttype is not defined; 
rewrite as a function with a return from every case/default

IvanFesenko (approved):

PR #1539, opened 2023-03-09
solution

vlad-trundenko:
[DEMO LINK](https://vlad-trundenko.github.io/react_list-of-goods/)


vlad-trundenko:
I accidentally completed an old version of the task, which I forked a long time ago(((

polosanya (changes requested):
Can you please delete a repository and fork a new one? The tasks are almost the same. 
Also, it will be good practice for you 🙂

PR #1551, opened 2023-03-17
Develop

Zelinskyi-Serhii:
[DEMO LINK](https://Zelinskyi-Serhii.github.io/react_list-of-goods/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1282,6 +1282,12 @@
             "psl": "^1.1.28",
             "punycode": "^2.1.1"
           }
+        },
+        "uuid": {
+          "version": "8.3.2",
+          "resolved": "https://registry.npmjs.org/uuid/-/uuid-8.3.2.tgz",
+          "integrity": "sha512-+NYs2QeMWy+GWFOEm9xnn6HCDp0l7QBD7ml8zLUmJ+93Q5NF0NocErnwkTkXVFNiX3/fpC6afS8Dhb/gz7R7eg==",
+          "dev": true
         }
       }
     },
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -26,70 +20,139 @@ type ReorderOptions = {
...
+  sortType: SortType,
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
+
+  reverseButton = () => {
```
rashyd-hasratov, line 60: I would recommend you to choose a better name for the method, 'reverseButton' sounds like the method reverses some button, not a list. The same is for 'resetedButton'

src/App.tsx
```diff
@@ -26,70 +20,139 @@ type ReorderOptions = {
...
+
+  sortByLength = () => {
+    this.setState({
+      sortType: SortType.LENGTH,
+    });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    const isReseted = (sortType !== SortType.NONE || isReversed);
```
rashyd-hasratov, line 87: I think it is better to change the name of the variable or rewrite the condition. `isReseted` means that the list is in original order, but the condition returns true if the list was reordered

```suggestion
    const isReordered = (sortType !== SortType.NONE || isReversed);
```

src/GoodList.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+type Props = {
+  good: string;
+};
+
+export class GoodList extends React.PureComponent<Props, {}> {
```
rashyd-hasratov, line 7: The class is called GoodList, but actually it returns `li` that is just list item. I would fix it

rashyd-hasratov (changes requested):
Good job! And it can be even better, check my suggestions please

src/GoodList.tsx
Zelinskyi-Serhii: I have some problems with naming:))

package-lock.json
OleksiiBlezniuk, line 1: don't add such files to PR

src/App.tsx
```diff
@@ -26,70 +20,139 @@ type ReorderOptions = {
...
+            className={classNames(
+              'button is-success',
+              {
+                'is-light': sortType !== SortType.LENGTH,
+              },
+            )}
+            onClick={this.sortByLength}
+          >
+            Sort by length
+          </button>
```
OleksiiBlezniuk, line 123: Would be nice to create an array from `sortType` enum (using Object.values) and map it to avoid code duplicating

src/App.tsx
```diff
@@ -26,70 +20,139 @@ type ReorderOptions = {
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {goods.map((good) => <Good key={uuidv4()} good={good} />)}
+          </ul>
```
OleksiiBlezniuk, line 153: As for me will be more logical to create component for the list and then move item to component if needed

OleksiiBlezniuk (approved):

PR #1554, opened 2023-03-18
Develop

Dynamizx:
https://dynamizx.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -59,12 +59,8 @@ export class App extends Component {
     isReversed: false,
   };
 
-  sortByAlphabet = () => {
-    this.setState({ sortType: SortType.ALPHABET });
-  };
-
-  sortByLength = () => {
-    this.setState({ sortType: SortType.LENGTH });
+  sortGoods = (sortType: SortType) => {
+    this.setState({ sortType });
   };
 
   reverse = () => {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,81 +15,114 @@ export const goodsFromServer = [
...
+
+const getSortFunction = (sortType: SortType): SortFunction => {
+  switch (sortType) {
+    case SortType.ALPHABET:
+      return (a, b) => a.localeCompare(b);
+    case SortType.LENGTH:
+      return (a, b) => a.length - b.length;
+    case SortType.NONE:
+    default:
+      return () => 0;
```
Anastasiia-Svintsova, line 39:
```suggestion
    default:
      return 0;
```

src/App.tsx
```diff
@@ -15,81 +15,114 @@ export const goodsFromServer = [
...
 
-        <button
-          type="button"
-          className="button is-success is-light"
-        >
-          Sort by length
-        </button>
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
```
Anastasiia-Svintsova, line 68: Can't you combine these 2 functions? You can pass a sort type as an argument

Anastasiia-Svintsova (changes requested):
- The `Reverse` button has active styles after first loading. Fix it to pass the test, please
image

https://user-images.githubusercontent.com/99403198/226101965-5692b9a1-f6ce-4ebf-8f54-a9437822200b.mov

src/App.tsx
```diff
@@ -15,81 +15,110 @@ export const goodsFromServer = [
...
+            Sort by length
+          </button>
+          <button
+            type="button"
+            className={`button is-warning ${this.state.isReversed ? '' : 'is-light'}`}
+            onClick={this.reverse}
+          >
+            Reverse
+          </button>
+          {(this.state.sortType !== SortType.NONE || this.state.isReversed) && (
```
polosanya, line 106: It would be great to create a variable for these values

polosanya (approved):
Good job!

PR #1556, opened 2023-03-18
Develop

OksanaShvets1:
[DEMO LINK](https://OksanaShvets1.github.io/react_list-of-goods/) and add it to the PR description.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,7 +2,7 @@ import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 import classNames from 'classnames';
-import { GoodsList } from './components/GoodsList';
+import { GoodsList } from './components/GoodsList/GoodsList';
 
 export const goodsFromServer = [
   'Dumplings',
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+import { GoodsList } from './components/GoodsList';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -21,75 +23,112 @@ enum SortType {
...
+  reset = () => {
+    this.setState({ sortType: SortType.NONE });
+  };
+
+  render() {
+    const { goods, isReversed, sortType } = this.state;
+
+    const visibleGoods = [...goods];
+
+    visibleGoods.sort((g1, g2) => {
```
TarasHoliuk, line 62: it's better to do it (and reverse) list in a separate function (that was declared initially - `getReorderedGoods`) and just call it in component (but outside the `render` method) to get `visibleGoods`

src/App.tsx
```diff
@@ -21,75 +23,112 @@ enum SortType {
...
-  );
-};
+        case SortType.LENGTH:
+          return g1.length - g2.length;
+
+        case SortType.NONE:
+          return 0;
+
+        default:
+          return 0;
```
TarasHoliuk, line 74:
```suggestion
        case SortType.NONE:
        default:
          return 0;
```

src/App.tsx
```diff
@@ -21,75 +23,112 @@ enum SortType {
...
-// Use this function in the render method to prepare goods
-export function getReorderedGoods(
-  goods: string[],
-  { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
+export class App extends Component <{}, State> {
+  state: State = {
+    goods: goodsFromServer,
```
TarasHoliuk, line 34: you don't need to save goods in the state

src/App.tsx
```diff
@@ -21,75 +23,112 @@ enum SortType {
...
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  reset = () => {
```
TarasHoliuk, line 53: it should reset `isReversed` as well

TarasHoliuk (changes requested):

tetiana-ivanchenko (approved):
GJ

PR #1559, opened 2023-03-19
listOfGoods was implement

KuharchukDmytro:
https://KuharchukDmytro.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,11 @@
-import React from 'react';
+import classNames from 'classnames';
+
+import { Component } from 'react';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+import { GoodsList } from './components/GoodsList';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


kateryna-babina (approved):

src/App.tsx
```diff
@@ -26,70 +31,116 @@ type ReorderOptions = {
...
+      );
+      break;
 
-  return visibleGoods;
-}
+    case SortType.LENGTH:
+      visibleGoods.sort(
+        (good1, good2) => good1.length - good2.length,
+      );
+      break;
```
OleksiiBlezniuk, line 51: But you've duplicated sort method, let's move switch/case inside sort method instead

src/App.tsx
```diff
@@ -26,70 +31,116 @@ type ReorderOptions = {
...
+          <button
+            type="button"
+            className={classNames(
+              'button is-success',
+              { 'is-light': sortType !== SortType.LENGTH },
+            )}
+            onClick={() => this.handleSort(SortType.LENGTH)}
+          >
+            Sort by length
+          </button>
```
OleksiiBlezniuk, line 118: Would be nice to create an array from `SortType` enum (use Object.values) and map it here instead of code repeating

OleksiiBlezniuk (approved):

PR #1567, opened 2023-03-20
solution

nastychenko:
https://nastychenko.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -42,54 +43,114 @@ export function getReorderedGoods(
...
+            onClick={() => {
+              this.setState(state => ({
+                isReversed: !state.isReversed,
+              }));
+            }}
+          >
+            Reverse
+          </button>
+
+          {(isReversed || sortType !== SortType.NONE) && (
```
polosanya, line 131: It would be great to save all these booleans into some variable, so it's clear why you need them

polosanya (changes requested):
Almost perfect!

polosanya (approved):
Well done!

PR #1580, opened 2023-03-21
Implemented list of goods with react

YaroslavDzyadiv:
[DEMO LINK](https://YarosalvDzyadiv.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,134 @@ type ReorderOptions = {
...
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {visibleGoods.map((good) => (
+            <li data-cy="Good" key={good}>
+              {good}
+            </li>
+          ))}
         </ul>
```
kuzmi491, line 156: Maybe it's better to create a separate component for this list? But is just my opinion=))

kuzmi491 (approved):

sergii-nosachenko (changes requested):
Fix demo link
Screenshot 2023-03-23 at 11 24 29

src/App.tsx
```diff
@@ -26,70 +27,134 @@ type ReorderOptions = {
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {visibleGoods.map((good) => (
+            <li data-cy="Good" key={good}>
+              {good}
+            </li>
+          ))}
```
OleksiiBlezniuk, line 155: Let's move goods list to separate component

src/App.tsx
```diff
@@ -26,70 +27,134 @@ type ReorderOptions = {
...
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
+    const showReset = sortType !== SortType.NONE || isReversed;
```
OleksiiBlezniuk, line 95:
```suggestion
    const shouldShowReset = sortType !== SortType.NONE || isReversed;
```

OleksiiBlezniuk: showReset - action that will show reset (button in this context)
shouldShowReset - defines if this action (render of this button) should be performed

src/App.tsx
```diff
@@ -26,70 +27,134 @@ type ReorderOptions = {
   isReversed: boolean,
 };
 
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
```
sergii-nosachenko, line 33: Add return type

src/App.tsx
```diff
@@ -26,70 +27,134 @@ type ReorderOptions = {
...
-          Reset
-        </button>
-      </div>
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+};
+
+export class App extends Component<{}, State> {
+  state = {
```
sergii-nosachenko, line 62: Make it Readonly

OleksiiBlezniuk (approved):
Will approve, but such things like lists of some products should be in separate components

In this case it's not critical, but in real project - in most cases this 7 lines of code easily can be transformed to 200 lines of component in few weeks (or days), so better to split logic to components from the beginning (why do something tomorrow if we can do it today?)

src/App.tsx
```diff
@@ -26,70 +27,134 @@ type ReorderOptions = {
...
+    sortType: SortType.NONE,
+  };
+
+  reverse = () => {
+    this.setState(state => ({
+      isReversed: !state.isReversed,
+    }));
+  };
+
+  sortTypeLength = () => {
```
sergii-nosachenko, line 73:
```suggestion
  sortByLength = () => {
```

src/App.tsx
```diff
@@ -26,70 +27,134 @@ type ReorderOptions = {
...
+    this.setState({
+      sortType: SortType.LENGTH,
+    });
+  };
+
+  sortTypeAlphabet = () => {
```
sergii-nosachenko, line 79: Same here

src/App.tsx
sergii-nosachenko, line 1: Add values to enum

src/App.tsx
sergii-nosachenko: @kuzmi491 Good point.
@YaroslavDzyadiv Fix it.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1595, opened 2023-03-22
list of goods react solution

ksavelyeva:
[DEMO LINK](https://ksavelyeva.github.io/react_list-of-goods/)

### DIFF
--- a/src/components/List.tsx
+++ b/src/components/List.tsx
@@ -1,0 +1,18 @@
+import React from 'react';
+
+type Props = {
+  goods: string[],
+};
+
+export const List: React.FC<Props> = ({ goods }) => (
+  <ul>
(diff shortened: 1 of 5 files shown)


src/components/List.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+type Props = {
+  goods: string[],
+};
+
+export const List: React.FC<Props> = ({ goods }) => (
```
OleksiiBlezniuk, line 7:
```suggestion
export const GoodsList: React.FC<Props> = ({ goods }) => (
```

OleksiiBlezniuk: This list has specific `goods` prop, so better to specify it in component name

PR #1612, opened 2023-03-29
react_list-of-goods was done

pushkarskiyrodion:
[DEMO LINK](https://pushkarskiyrodion.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
@@ -26,7 +28,6 @@ type ReorderOptions = {
   isReversed: boolean,
 };
 
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +28,119 @@ type ReorderOptions = {
...
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((g1, g2) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return g1.localeCompare(g2);
+      case SortType.LENGTH:
+        return g1.length - g2.length;
+      default:
+        return 0;
```
OlEzhA0, line 45: add empty lines between blocks
```suggestion
      case SortType.ALPHABET:
        return g1.localeCompare(g2);

      case SortType.LENGTH:
        return g1.length - g2.length;

      default:
        return 0;
```

src/App.tsx
```diff
@@ -26,70 +28,125 @@ type ReorderOptions = {
...
+
+          <button
+            type="button"
+            className={cn('button is-warning', {
+              'is-light': isReversed === false,
+            })}
+            onClick={() => {
+              this.setState((state) => ({
+                isReversed: !state.isReversed,
+              }));
```
Volodymyr-Melnyk-Fron10der, line 113: there is no a lot of logic, but I would move the code into function if it takes more 2 lines

Volodymyr-Melnyk-Fron10der (approved):
Well done✅

PR #1618, opened 2023-04-03
commit message

diamat1997:
[DEMO LINK](https://diamat1997.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { PureComponent } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,113 @@ type ReorderOptions = {
...
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  visibleGoods.sort((f1, f2) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return f1.localeCompare(f2);
```
IvanFesenko, line 39: No need run sort function in case when sortType is NONE

IvanFesenko (approved):
Check comments to avoid same mistakes in the future

PR #1620, opened 2023-04-06
solution :)

katerynashylina:
DEMO LINK](https://katerynashylina.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


IvanFesenko (changes requested):
Just small improvements needs to

src/App.tsx
```diff
@@ -31,65 +32,127 @@ export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  if (!SortType.NONE) {
```
etojeDenys, line 37: The condition `!SortType.NONE` will always be true, as `SortType.NONE` is 0 and `!0` is true. Instead of checking for `!SortType.NONE`, check for `sortType !== SortType.NONE`.

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -31,65 +32,127 @@ export function getReorderedGoods(
...
-  console.log(sortType, isReversed);
+  if (sortType !== SortType.NONE) {
+    visibleGoods.sort((g1, g2): number => {
+      switch (sortType) {
+        case SortType.ALPHABET:
+          return g1.localeCompare(g2);
+        case SortType.LENGTH:
+          return g1.length - g2.length;
+        default:
+          return 0;
```
loralevitska, line 45:
```suggestion
        case SortType.ALPHABET:
          return g1.localeCompare(g2);
          
        case SortType.LENGTH:
          return g1.length - g2.length;
          
        default:
          return 0;
```

loralevitska (approved):

PR #1622, opened 2023-04-07
Solution

DaveBeetle:
[DEMO LINK](https://DaveBeetle.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+          return good1.localeCompare(good2);
+
+        case SortType.LENGTH:
+          return good1.length - good2.length;
+
+        default:
+          return 0;
+      }
+    });
+  }
```
IvanFesenko, line 49: No need to sort if sort type NONE. Just fix condition `if (sortType !== SortType.NONE)`

IvanFesenko (changes requested):
On small improvement required

Volodymyr-Melnyk-Fron10der (approved):
WellDoneGIF

PR #1623, opened 2023-04-07
Develop

nazar-medushevskyi:
- [DEMO LINK](https://nazar-medushevskyi.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


anstsot (changes requested):
Looks good but there is one style issue. Buttons seems to be active by default and don't change the color on click. could you fix it?
image

src/App.tsx
```diff
@@ -26,69 +26,110 @@ type ReorderOptions = {
...
-  console.log(sortType, isReversed);
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort();
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+    default:
+      break;
```
loralevitska, line 43:
```suggestion
    case SortType.ALPHABET:
      visibleGoods.sort();
      break;
      
    case SortType.LENGTH:
      visibleGoods.sort((a, b) => a.length - b.length);
      break;
      
    default:
      break;
```

src/App.tsx
```diff
@@ -26,69 +26,110 @@ type ReorderOptions = {
...
+
+  const visibleGoods = getReorderedGoods(goodsFromServer, reorderOptions);
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={`button is-info is-light ${reorderOptions.sortType === SortType.ALPHABET ? 'is-active' : ''}`}
```
loralevitska, line 94: use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally

loralevitska (approved):

PR #1625, opened 2023-04-07
Solution

yao157:
[DEMO LINK](https://yao157.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,125 @@ type ReorderOptions = {
...
+      case SortType.ALPHABET:
+        return a.localeCompare(b);
+      case SortType.LENGTH:
+        return a.length - b.length;
+      case SortType.NONE:
+        return 0;
+      default:
+        throw new Error('Wrong sort type');
+    }
+  });
```
DarkMistyRoom, line 47: would be better for the performance to set callback outside. Because now we do unnecessary iterations in NONE case

DarkMistyRoom (approved):
Good job!

PR #1631, opened 2023-04-12
solution

ladrom:
[demo](https://ladrom.github.io/react_list-of-goods/)


src/App.tsx
```diff
@@ -26,70 +27,136 @@ type ReorderOptions = {
...
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+export class App extends React.Component {
+  state: ReorderOptions = {
+    sortType: 0,
```
agniya-i, line 64:
```suggestion
    sortType: SortType.NONE
```

use enum `SortType` for `sortType`

src/App.tsx
```diff
@@ -26,70 +27,136 @@ type ReorderOptions = {
...
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            onClick={this.alphabeticalBtn}
+            type="button"
+            className={classNames(
+              'button',
+              'is-info',
+              { 'is-light': sortType !== 1 },
```
agniya-i, line 103:
```suggestion
              { 'is-light': sortType !== SortType.ALPHABET },
```

fix it everywhere

agniya-i (changes requested):

src/App.tsx
ladrom: Thanks for the help.

src/App.tsx
```diff
@@ -26,70 +27,139 @@ type ReorderOptions = {
...
-  console.log(sortType, isReversed);
+  switch (sortType) {
+    case 1:
+      visibleGoods.sort((g1, g2) => {
+        return g1.localeCompare(g2);
+      });
+
+      break;
+
+    case 2:
```
polosanya, line 44: Instead of using numbers as case values in the switch statement, use the SortType enum for better readability and maintainability

src/App.tsx
```diff
@@ -26,70 +27,139 @@ type ReorderOptions = {
...
+      });
+
+      break;
+
+    default:
+      break;
+  }
+
+  if (isReversed) {
+    return [...visibleGoods].reverse();
```
polosanya, line 56: No need to make one more copy, as visibleGoods is a copy of goods

src/App.tsx
```diff
@@ -26,70 +27,139 @@ type ReorderOptions = {
...
+            type="button"
+            className={classNames(
+              'button',
+              'is-danger',
+              'is-light',
+              { 'is-hidden': sortType === SortType.NONE && !isReversed },
+            )}
+          >
+            Reset
+          </button>
```
polosanya, line 147: Don't make it hidden. Just render this button only if sortType is different from none or goods were reversed to pass the tests. Use conditional rendering for that
https://react.dev/learn/conditional-rendering

polosanya (changes requested):
Almost done!

anstsot (approved):
Great job, approved!

PR #1635, opened 2023-04-14
solution

PashaMalyshkin:
[DEMO LINK](https://PashaMalyshkin.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
@@ -16,9 +17,9 @@ export const goodsFromServer = [
 ];
 
 enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
+  NONE = 'none',
+  ALPHABET = 'alphabet',
+  LENGTH = 'length',
 }
 
 type ReorderOptions = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -16,80 +17,125 @@ export const goodsFromServer = [
...
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+};
```
Esteban4769, line 59: You have type `ReorderOptions` already, u dont need this.

src/App.tsx
```diff
@@ -16,80 +17,125 @@ export const goodsFromServer = [
 ];
 
 enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
+  NONE = 'none',
+  ALPHABET = 'alphabet',
+  LENGTH = 'length',
```
Esteban4769, line 22: Dont really need string literals.

src/App.tsx
```diff
@@ -16,80 +17,125 @@ export const goodsFromServer = [
...
+  if (sortType !== 'none') {
+    visibleGoods.sort((firstGood, secondGood) => {
+      switch (sortType) {
+        case 'alphabet':
+          return firstGood.localeCompare(secondGood);
+        case 'length':
+          return firstGood.length - secondGood.length;
+        default:
+          return 0;
+      }
```
Esteban4769, line 45: Kinda weird code. I would sort inside `switch-case`, removed condition `(sortType !== 'none')`  and add case 'none'. Also in default case, u should throw an error.

Esteban4769 (approved):
Good job, can make some adjusment if u want.

src/App.tsx
PashaMalyshkin: Agree, thanks

src/App.tsx
PashaMalyshkin: agree

src/App.tsx
PashaMalyshkin: Yeah, ty

src/App.tsx
```diff
@@ -26,70 +27,108 @@ type ReorderOptions = {
...
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  resetGoods = () => {
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
```
liliia-livandovskaya, line 79: you can combine this into a single function with switch/case

PR #1637, opened 2023-04-16
Solution

emxm:


emxm:
https://emxm.github.io/react_list-of-goods/

src/App.tsx
```diff
@@ -34,62 +35,137 @@ export function getReorderedGoods(
...
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+
+    default:
```
Volodymyr-Melnyk-Fron10der, line 47: return just visible goods in default

Volodymyr-Melnyk-Fron10der (changes requested):
1. Reverse button is working with some bugs, fix it. There is some point when it's just not being able to select

src/App.tsx
vlad-demchuk: https://github.com/mate-academy/react_list-of-goods/pull/1637/files#r1167989440
not fixed

vlad-demchuk (changes requested):

Anastasiia-Svintsova (approved):

PR #1638, opened 2023-04-16
Implemented the solution

zynkd:
[DEMO LINK](https://zynkd.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -22,74 +23,120 @@ enum SortType {
...
+      isReversed: !state.isReversed,
+    }));
+  };
+
+  // Revert back to the initial (default) state
+  handleReset = () => {
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    });
```
ivanchuk-yurii, line 79: it's better to create variable for default state and use it there

src/App.tsx
```diff
@@ -22,74 +23,120 @@ enum SortType {
...
+            type="button"
+            className={classNames('button', 'is-warning', {
+              'is-light': isReversed === false,
+            })}
+            onClick={this.handleReverse}
+          >
+            Reverse
+          </button>
+
+          {(sortType === SortType.NONE && !isReversed) || (
```
ivanchuk-yurii, line 121: also compare here with default state

ivanchuk-yurii (approved):

PR #1639, opened 2023-04-16
list-of-goods sollution

serbinm:
[DEMO LINK](https://serbinm.github.io/react_list-of-goods/)


polosanya (changes requested):
Almost done!

PR #1642, opened 2023-04-18
made list of goods

PavloReutskiy:
[DEMO LINK](https://PavloReutskiy.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
+import classNames from 'classnames';
 import './App.scss';
 
 export const goodsFromServer = [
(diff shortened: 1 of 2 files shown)


AnyaLyaLya (approved):
GJ!

src/App.tsx
```diff
@@ -26,70 +27,129 @@ type ReorderOptions = {
...
+
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  reset = () => {
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
```
liliia-livandovskaya, line 66: you can combine this into a single function with switch/case

PavloReutskiy: I'm not sure I did it correctly, but I did it using switch / case

PR #1652, opened 2023-04-19
list of goods

danylo-dobrovolskyi:
[DEMO LINK](https://danylo-dobrovolskyi.github.io/react_list-of-goods/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Implement the ability to reorder the goods with the given buttons.
     - remove `is-light` class from the button when this order is applied;
 1. `Sort by length` button should sort the goods by length.
     - remove `is-light` class from the button when this order is applied;
-1. `Reverse` button changes the current order to the opposite one. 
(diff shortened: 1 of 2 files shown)


makandriival (changes requested):
[
Screenshot 2023-04-20 at 15 10 25
]

something wrong in the demo, sort by length button specifically. Functionality works but the visual is not changing as expected.

also for class management its better to use ClassNames library

danylo-dobrovolskyi:
> 



> [ Screenshot 2023-04-20 at 15 10 25 ]
> 
> something wrong in the demo, sort by length button specifically. Functionality works but the visual is not changing as expected.
> 
> also for class management its better to use ClassNames library

> 

Thanks for your suggestions! The problem was in one missing letter in className of 'Sort by length' button, also I added ClassNames library

SerhiiKirik (approved):
Great 🔥

PR #1658, opened 2023-04-20
Solution

Zad0rnik:
[DEMO LINK](https://Zad0rnik.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,70 +26,119 @@ type ReorderOptions = {
   isReversed: boolean,
 };
 
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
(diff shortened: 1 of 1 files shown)


yuliia-anisimova (approved):
GJ!

src/App.tsx
```diff
@@ -26,70 +26,119 @@ type ReorderOptions = {
...
+      sortType: SortType.LENGTH,
+    });
+  };
+
+  reset = () => {
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
```
liliia-livandovskaya, line 82: you can combine this into a single function with switch/case

src/App.tsx
```diff
@@ -26,70 +26,119 @@ type ReorderOptions = {
...
+
+  render() {
+    const { sortType, isReversed } = this.state;
+    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
```
liliia-livandovskaya, line 86: Instead of passing the entire this.state to getReorderedGoods(), you should pass only the required properties, like { sortType, isReversed }.

liliia-livandovskaya (approved):

PR #1666, opened 2023-04-20
Solution

Kavon-Skela:
https://Kavon-Skela.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -38,58 +39,124 @@ export function getReorderedGoods(
...
+  render() {
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={classNames(
+              'button',
+              'is-info',
+              { 'is-light': this.state.sortType !== SortType.ALPHABET },
```
Goto-Jesus, line 109: You can destructuring your variables
Like this `const { isReversed, sortType } = this.state;`
for not writing `this.state.sortType.....`

TanyaS133 (approved):

src/App.tsx
```diff
@@ -38,58 +39,124 @@ export function getReorderedGoods(
...
+              'is-warning',
+              { 'is-light': !this.state.isReversed },
+            )}
+            onClick={this.reverseGoods}
+          >
+            Reverse
+          </button>
+
+          {(this.state.isReversed !== false
+          || this.state.sortType !== SortType.NONE)
```
Goto-Jesus, line 141: the condition better like this:
`(sortType !== SortType.NONE || isReversed)`

than this:
`this.state.isReversed !== false`
really strange 😅

Goto-Jesus (changes requested):
please consider my messages and make corrections

src/App.tsx
Kavon-Skela: It could be extremely strange, but it is my decision. You should be more polite with your words, cause your words to sound a bit offensive

src/App.tsx
```diff
@@ -38,58 +39,125 @@ export function getReorderedGoods(
...
+      };
+    });
+  };
+
+  resetAllSorts = () => {
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    });
+  };
```
liliia-livandovskaya, line 98: you can combine this into a single function with switch/case

src/App.tsx
```diff
@@ -38,58 +39,125 @@ export function getReorderedGoods(
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
```
liliia-livandovskaya, line 40: you can remove it

liliia-livandovskaya (approved):

src/App.tsx
Goto-Jesus: i am so sorry i didn't want to offend you (((

Kavon-Skela: Ok, np)

src/App.tsx
Kavon-Skela: I didn't get the main idea. Did u mean getReorderedGoods function?

PR #1673, opened 2023-04-20
List of goods

roshirin:
[DEMO LINK](https://roshirin.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+  };
+
+  handleReset = () => {
+    this.setState({
+      sortType: 0,
+      isReversed: false,
+    });
+  };
+
+  sortByAlphabet = () => {
```
yuliia-anisimova, line 83: I think you should change the method's name here, so it starts with the verb 'handle' according to the checklist (check p.4 GGOD EXAMPLE)

src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+    this.setState({ sortType: 1 });
+  };
+
+  sortByLength = () => {
```
yuliia-anisimova, line 87: The same problem with the method's name here

src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
+  visibleGoods.sort((a, b) => {
```
yuliia-anisimova, line 36: I think it's better to avoid using names like 'a' or 'b', because you actually sort goods here

yuliia-anisimova (changes requested):
Awesome! But please, look through my comments

src/App.tsx
roshirin: Function name should start with a verb, but not actualy "handle". "Sort" is a verb so I think it fits just fine.

src/App.tsx
roshirin: Sort is always used with (a, b) parameters, it's a kind of commonplace. I would be more confused with the names like 'good1' and 'good2'. It would be less readable in my opinion, but I'm ready to change the names if menthor says so.

src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+    const resetButtonVisible = isReversed || sortType !== 0;
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={
+              classNames('button is-info', {
+                'is-light': sortType !== 1,
```
SerhiiKirik, line 103: use Enum for it

src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+              classNames('button is-warning', {
+                'is-light': !isReversed,
+              })
+            }
+            onClick={this.handleReverse}
+          >
+            Reverse
+          </button>
+
+          {resetButtonVisible && (
```
SerhiiKirik, line 135:
```suggestion
          {IsResetButtonVisible && (
```

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your solution 🧑‍💻

src/App.tsx
```diff
@@ -26,70 +27,132 @@ type ReorderOptions = {
...
+  };
+
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    const goods = getReorderedGoods(goodsFromServer, { isReversed, sortType });
+    const isResetButtonVisible = isReversed || sortType !== SortType.NONE;
```
jstmpelowycz, line 94:
```suggestion
    const { isReversed, sortType } = this.state;

    const goods = getReorderedGoods(goodsFromServer, { 
      isReversed,
      sortType,
    });
    
    const isResetButtonVisible = isReversed || sortType !== SortType.NONE;
```

jstmpelowycz (approved):

PR #1674, opened 2023-04-20
Solution

MylaCrystal:
[DEMO LINK](https://MylaCrystal.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +26,131 @@ type ReorderOptions = {
...
+      isReversed: !state.isReversed,
+    }));
+  };
+
+  reset = () => {
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    });
+  };
```
kukla1989, line 92: 4. [CODE STYLE] - Method's name should start with a verb (so you could easily tell what your method actually do)
its better that such type of methods start with verb "handler", in my code i declare like "handleResetClick" and "handleSortLengthClick"

kukla1989 (changes requested):
good job! just do minor renaming

src/App.tsx
MylaCrystal: I'll rename them using the "handle" prefix, but in general, I don't think this checklist item is appropriate here since reverse & reset is essentially a verb

src/App.tsx
```diff
@@ -26,70 +26,131 @@ type ReorderOptions = {
...
+  visibleGoods.sort((good1, good2) => {
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
+
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      default:
+        return 0;
```
jstmpelowycz, line 44:
```suggestion
      case SortType.NONE:
      default:
        return 0;
```

src/App.tsx
```diff
@@ -26,70 +26,131 @@ type ReorderOptions = {
...
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
```
jstmpelowycz, line 53:
```suggestion
```

src/App.tsx
```diff
@@ -26,70 +26,131 @@ type ReorderOptions = {
...
+    const { isReversed, sortType } = this.state;
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={sortType === SortType.ALPHABET
+              ? 'button is-info'
+              : 'button is-info is-light'}
```
jstmpelowycz, line 105: let's use classnames

src/App.tsx
```diff
@@ -26,70 +26,131 @@ type ReorderOptions = {
...
 
-      <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
```
jstmpelowycz, line 145: extra `ul`

jstmpelowycz (changes requested):

src/App.tsx
```diff
@@ -26,70 +27,131 @@ type ReorderOptions = {
...
+            onClick={this.handleSortByAlphabet}
+          >
+            Sort alphabetically
+          </button>
+
+          <button
+            type="button"
+            className={sortType === SortType.LENGTH
+              ? 'button is-primary'
+              : 'button is-success is-light'}
```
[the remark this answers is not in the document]
MylaCrystal: oops 😅
sorry, my bad 😅

src/App.tsx
```diff
@@ -26,70 +27,127 @@ type ReorderOptions = {
...
+            type="button"
+            className={cn('button', 'is-warning', {
+              'is-light': !isReversed,
+            })}
+            onClick={this.handleReverseItems}
+          >
+            Reverse
+          </button>
+
+          {(sortType !== SortType.NONE || isReversed) && (
```
SerhiiKirik, line 130: move this check to the separate variable
```suggestion
          {... && (
```

SerhiiKirik (approved):
Great 🔥

PR #1699, opened 2023-05-02
Solution

LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +26,103 @@ type ReorderOptions = {
...
+    case LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+    case ALPHABET:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+    case NONE:
+      visibleGoods = [...goods];
+      break;
+    default:
```
witflash, line 46: You can do `break` on default:
```
default:
  break;
```
It looks better.

src/App.tsx
```diff
@@ -26,70 +26,103 @@ type ReorderOptions = {
...
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+
+    const goods = getReorderedGoods(goodsFromServer, { sortType, isReversed });
+    const { ALPHABET, LENGTH, NONE } = SortType;
+
+    return (
+      <>
```
witflash, line 77: You don't need this Fragment. You already have `div` as the root wrapper.

src/App.tsx
```diff
@@ -26,70 +26,103 @@ type ReorderOptions = {
...
+              onClick={() => this.setState({ sortType: ALPHABET })}
+              type="button"
+              className={`button is-info ${sortType === ALPHABET ? '' : 'is-light'}`}
+            >
+              Sort alphabetically
+            </button>
+
+            <button
+              type="button"
+              className={`button is-success ${sortType === LENGTH ? '' : 'is-light'}`}
```
witflash, line 90: You can use some libs (for example `classnames`) to use conditional class names. It may be more readable.

witflash (approved):
Well done 👍 
Pay attention to my comments for your future work.

src/App.tsx
LBondarchuk: my personal opinion is that it is better and clearer
I know how to use this library

PR #1702, opened 2023-05-06
Solution

andrii-kovalskyi02:
[DEMO LINK](https://andrii-kovalskyi02.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,106 @@ type ReorderOptions = {
...
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  switch (sortType) {
+    case 1:
```
TarasHoliuk, line 37: It's not clear what is `1` here - don't use an enum value in this way. 
```suggestion
    case SortType.ALPHABET:
```
fix it everywhere

TarasHoliuk (changes requested):

SanyaBratashchuk (approved):

PR #1709, opened 2023-05-09
solution

MikiOOOO:
https://MikiOOOO.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -16,78 +16,119 @@ export const goodsFromServer = [
...
+      throw new Error('invalid sortType key');
+    }
+  }
+
+  return isReversed ? visibleGoods.reverse() : visibleGoods;
+}
 
 export const App: React.FC = () => {
+  const [isReversed, setIsReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
```
danielpopek94, line 64: I think you should not use hooks in this exercise.

danielpopek94 (changes requested):

src/App.tsx
```diff
@@ -16,80 +16,128 @@ export const goodsFromServer = [
 ];
 
 enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
+  NONE = 0,
+  ALPHABET = 1,
+  LENGTH = 2,
 }
```
d-wasilewski, line 22: you can leave this part as it was

src/App.tsx
```diff
@@ -16,80 +16,128 @@ export const goodsFromServer = [
...
 
-type ReorderOptions = {
+interface State {
   sortType: SortType,
-  isReversed: boolean,
-};
+  isReversed: boolean | null,
```
d-wasilewski, line 26: this should always be either true or false, never null

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1717, opened 2023-05-12
solution v1

bjeziorek:
[DEMO LINK](https://bjeziorek.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,8 +22,8 @@ enum SortType {
 }
 
 type ReorderOptions = {
-  sortType: SortType,
-  isReversed: boolean,
+  sortType: SortType;
+  isReversed: boolean;
 };
 
 // Use this function in the render method to prepare goods
@@ -32,9 +32,25 @@ export function getReorderedGoods(
   { sortType, isReversed }: ReorderOptions,
 ) {
   // To avoid the original array mutation
-  const visibleGoods = [...goods];
+  let visibleGoods = [...goods];
 
   // Sort and reverse goods if needed
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods
+      = visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
+      break;
+    case SortType.LENGTH:
+      visibleGoods
+      = visibleGoods.sort((good1, good2) => good1.length - good2.length);
+      break;
+    default: // NONE
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -32,64 +32,97 @@ export function getReorderedGoods(
...
+    case SortType.ALPHABET:
+      visibleGoods
+      = visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
+      break;
+    case SortType.LENGTH:
+      visibleGoods
+      = visibleGoods.sort((good1, good2) => good1.length - good2.length);
+      break;
+    default: // NONE
+  }
```
kasiakubisiak, line 48: You can sort whole switch statement instead of doing it separately in each case, also use better naming:

```suggestion
  visibleGoods.sort((firstGood, secondGood) => {
    switch (sortType) {
      case SortType.ALPHABET:
        return firstGood.localeCompare(secondGood);
      case SortType.LENGTH:
        return firstGood.length - secondGood.length;
      default:
        return 0;
    }
  });
```

src/App.tsx
```diff
@@ -32,64 +32,97 @@ export function getReorderedGoods(
...
 
 // DON'T save goods to the state
-// type State = {
-//   isReversed: boolean,
-//   sortType: SortType,
-// };
+ type State = {
+   isReversed: boolean,
+   sortType: SortType,
+ };
```
kasiakubisiak, line 64: You don't need to use it at all
```suggestion
```

src/App.tsx
```diff
@@ -32,64 +32,97 @@ export function getReorderedGoods(
...
-  return (
-    <div className="section content">
-      <div className="buttons">
-        <button
-          type="button"
-          className="button is-info is-light"
-        >
-          Sort alphabetically
-        </button>
+export class App extends React.Component<{}, State> {
```
kasiakubisiak, line 66:
```suggestion
export class App extends React.Component<{}, ReorderOptions > {
```

src/App.tsx
```diff
@@ -32,64 +32,97 @@ export function getReorderedGoods(
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {getReorderedGoods(
+              goodsFromServer,
+              this.state,
+            ).map(good => <li data-cy="Good" key={good}>{good}</li>)}
```
kasiakubisiak, line 122:
```suggestion
          {visibleGoods.map(good => (<li data-cy="Good" key={good}>{good}</li>)}
```

kasiakubisiak (changes requested):

src/App.tsx
```diff
@@ -32,64 +32,97 @@ export function getReorderedGoods(
...
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods
+      = visibleGoods.sort((good1, good2) => good1.localeCompare(good2));
+      break;
+    case SortType.LENGTH:
+      visibleGoods
+      = visibleGoods.sort((good1, good2) => good1.length - good2.length);
+      break;
+    default: // NONE
```
kasiakubisiak, line 47:
```suggestion
      default:
        return 0;
```

kasiakubisiak (approved):

PR #1718, opened 2023-05-12
Solution

SuaBy514:
[DEMO LINK](https://suaby514.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -36,60 +36,107 @@ export function getReorderedGoods(
 
   // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  if (sortType !== SortType.NONE) {
+    visibleGoods.sort((firstGood, secondGood) => {
+      if (sortType === SortType.ALPHABET) {
+        return firstGood.localeCompare(secondGood);
+      }
+
+      if (sortType === SortType.LENGTH) {
+        return firstGood.length - secondGood.length;
+      }
+
+      return 0; // this is to make linter shut the hell up
+    });
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
 
   return visibleGoods;
 }
 
-// DON'T save goods to the state
-// type State = {
-//   isReversed: boolean,
-//   sortType: SortType,
-// };
-
-export const App: React.FC = () => {
-  return (
-    <div className="section content">
-      <div className="buttons">
-        <button
-          type="button"
-          className="button is-info is-light"
-        >
-          Sort alphabetically
-        </button>
-
-        <button
-          type="button"
-          className="button is-success is-light"
-        >
-          Sort by length
-        </button>
-
-        <button
-          type="button"
-          className="button is-warning is-light"
-        >
-          Reverse
-        </button>
-
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
-
-      <ul>
-        <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
-        </ul>
-      </ul>
-    </div>
-  );
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
 };
+
+export class App extends React.Component<{}, State> {
+  state = {
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
+
+  handleButton(actionType: string) {
+    if (actionType === 'reverse') {
+      this.setState((state) => ({ isReversed: !state.isReversed }));
+    }
+
+    if (actionType === 'reset') {
+      this.setState({ isReversed: false, sortType: SortType.NONE });
+    }
+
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -36,60 +36,107 @@ export function getReorderedGoods(
...
+        return firstGood.localeCompare(secondGood);
+      }
+
+      if (sortType === SortType.LENGTH) {
+        return firstGood.length - secondGood.length;
+      }
+
+      return 0; // this is to make linter shut the hell up
+    });
+  }
```
kasiakubisiak, line 51: maybe lets do switch/case instead of if statements

src/App.tsx
```diff
@@ -36,60 +36,107 @@ export function getReorderedGoods(
...
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+};
```
kasiakubisiak, line 63: you don't need to use State
```suggestion
```

src/App.tsx
```diff
@@ -36,60 +36,107 @@ export function getReorderedGoods(
...
+
+export class App extends React.Component<{}, State> {
```
kasiakubisiak, line 65:
```suggestion
export class App extends React.Component<{}, ReorderOptions> {
```

src/App.tsx
```diff
@@ -36,60 +36,107 @@ export function getReorderedGoods(
...
+    }
+
+    if (actionType === 'sortAlphabet') {
+      this.setState({ sortType: SortType.ALPHABET });
+    }
+
+    if (actionType === 'sortLength') {
+      this.setState({ sortType: SortType.LENGTH });
+    }
+  }
```
kasiakubisiak: maybe instead of so many if statements lets separate each of them to something like:
```
  reverseOrder = () => {
    this.setState(state => ({
    }));
  };
```
and then desctructure it and use it in button's onClick: onClick={reverseOrder}

src/App.tsx
```diff
@@ -36,60 +36,107 @@ export function getReorderedGoods(
...
-      <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {sortedGoods.map((good) => (
+            <li data-cy="Good">{good}</li>
```
kasiakubisiak, line 136: add key

src/App.tsx
```diff
@@ -36,60 +36,107 @@ export function getReorderedGoods(
...
 
-      <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {sortedGoods.map((good) => (
```
kasiakubisiak, line 135: you don't need to use `sortedGoods`, go for `visibleGoods`

kasiakubisiak (changes requested):

src/App.tsx
SuaBy514: This isn't the version with the hooks, why shouldn't I use state?

SuaBy514: Nevermind, I saw the comment below, I feel dumb now

src/App.tsx
SuaBy514: With onClick={reverseOrder} and the like, the list does not update
Using onClick={() => (reverseOrder())} instead works

src/App.tsx
```diff
@@ -36,60 +36,104 @@ export function getReorderedGoods(
 
   // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  if (sortType !== SortType.NONE) {
```
kasiakubisiak, line 39: you don't have to put switch case inside if statement

kasiakubisiak (changes requested):

src/App.tsx
```diff
@@ -36,60 +36,104 @@ export function getReorderedGoods(
...
+    const visibleGoods = getReorderedGoods(
+      goodsFromServer,
+      { sortType, isReversed },
+    );
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            onClick={() => (this.sortAlphabetically())}
```
kasiakubisiak, line 97: lets destructure 'this'

src/App.tsx
```diff
@@ -36,60 +36,104 @@ export function getReorderedGoods(
...
-          Reset
-        </button>
-      </div>
+export class App extends React.Component<{}, ReorderOptions> {
+  state = {
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
+
+  reset() {
```
kasiakubisiak, line 70: check checklist
```suggestion
  handleReset() {
```

kasiakubisiak (changes requested):

src/App.tsx
kasiakubisiak: Yes but let's do handleReset to make it indicate that a function is an event handler

kasiakubisiak (approved):

PR #1754, opened 2023-06-01
solution

devWizHub:
[DEMO LINK] https://devWizHub.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,75 +21,106 @@ enum SortType {
   LENGTH,
 }
 
-type ReorderOptions = {
-  sortType: SortType,
+type State = {
   isReversed: boolean,
+  sortType: SortType,
 };
 
-// Use this function in the render method to prepare goods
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -21,75 +21,106 @@ enum SortType {
...
+    const isOriginalOrder = !isReversed && sortType === SortType.NONE;
+
+    const reverseButtonClasses = `button is-warning ${isReversed ? '' : 'is-light'}`;
 
-      <ul>
+    if (sortType === SortType.ALPHABET) {
+      goodsToDisplay.sort();
+    } else if (sortType === SortType.LENGTH) {
+      goodsToDisplay.sort((a, b) => a.length - b.length);
+    }
```
Anastasiia-Svintsova, line 73: Would be better to call the sort function only once and use switch/case inside it

Anastasiia-Svintsova (approved):

PR #1760, opened 2023-06-10
solution

KseniyaGonchar:
https://KseniyaGonchar.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,70 +27,135 @@ type ReorderOptions = {
...
+  }
+
+  visibleGoods.sort((g1: string, g2: string) => {
+    if (sortType === SortType.ALPHABET && isReversed) {
+      return g2.localeCompare(g1);
+    }
+
+    if (sortType === SortType.LENGTH && isReversed) {
+      return g2.length - g1.length;
+    }
```
OleziO, line 47: Replace with switch case construction. And make it nested to` if(isReversed)`

OleziO (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #1764, opened 2023-06-11
Solution

Ssviatt:
[DEMO LINK](https://Ssviatt.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -48,47 +65,71 @@ export function getReorderedGoods(
 // };
 
 export const App: React.FC = () => {
+  const [isReversed, setReversed] = useState(false);
+  const [sortType, setSortType] = useState(SortType.NONE);
+
+  const reversed = () => {
+    setReversed(!isReversed);
```
loralevitska, line 72: Better use previous state, to be clear that we are using the previous state, such a record will be more readable
```suggestion
    setReversed(prevState => !prevState);
```

src/App.tsx
```diff
@@ -48,47 +65,71 @@ export function getReorderedGoods(
...
+    setSortType(SortType.ALPHABET);
+  };
+
   return (
     <div className="section content">
       <div className="buttons">
         <button
           type="button"
-          className="button is-info is-light"
+          className={`button is-info ${sortType === SortType.ALPHABET ? '' : 'is-light'}`}
```
loralevitska, line 93: Use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally. Fix all cases

loralevitska (changes requested):
Almost perfect) Let's improve (check my comments)

PR #1767, opened 2023-06-12
react_list-of-goods

YelyzavetaMolochynska:
[DEMO LINK](https://YelyzavetaMolochynska.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,116 @@ type ReorderOptions = {
...
+    switch (sortType) {
+      case SortType.ALPHABET:
+        return first.localeCompare(second);
+      case SortType.LENGTH:
+        return first.length - second.length;
+      default:
+        return 0;
+    }
+  });
+
```
IvanFesenko, line 46: No need to sort in case sortType is NONE

IvanFesenko (changes requested):
2023-06-12_16h30_37
Your implementation has always reset butoon but it chages only when reverse is active
https://github.com/mate-academy/react_list-of-goods/assets/61904995/05f72f3c-277a-4be3-b61d-64c6fb609491

VitaliyBondarenko1982 (approved):

PR #1770, opened 2023-06-13
react list of goods was added

Olena-Shevchenko:
[DEMO LINK](https://Olena-Shevchenko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,7 +26,6 @@ type ReorderOptions = {
   isReversed: boolean,
 };
 
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
@@ -56,16 +55,10 @@ export function getReorderedGoods(
   if (isReversed) {
     visibleGoods.reverse();
   }
-  // To avoid the original array mutation
-
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
 
   return visibleGoods;
 }
 
-// DON'T save goods to the state
 type State = {
   sortType: SortType,
   isReversed: boolean,
@@ -101,7 +94,7 @@ export class App extends React.Component<{}, State> {
   render() {
     const { isReversed, sortType } = this.state;
     const goods = getReorderedGoods(goodsFromServer, { sortType, isReversed });
-    const startOrder = sortType === SortType.NONE && isReversed === false;
+    const startOrder = sortType === SortType.NONE && !isReversed;
 
     return (
       <div className="section content">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -31,9 +31,33 @@ export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
+  switch (sortType) {
+    case 0:
```
polosanya, line 37: Use Enum instead of number values

src/App.tsx
```diff
@@ -42,54 +66,95 @@ export function getReorderedGoods(
...
+          <button
+            onClick={this.reverse}
+            type="button"
+            className={`button is-warning ${isReversed ? '' : ' is-light'}`}
+          >
+            Reverse
+          </button>
+          {sortType === 0 && isReversed === false
+            ? ''
+            : (
```
polosanya, line 133: It would be great to create a variable for that condition. Also, you can you logical && instead of the ternary operator as you only decide whether to show the button or not
https://react.dev/learn/conditional-rendering#logical-and-operator-

src/App.tsx
```diff
@@ -42,54 +66,95 @@ export function getReorderedGoods(
...
+              <button
+                onClick={this.reset}
+                type="button"
+                className="button is-danger is-light"
+              >
+                Reset
+              </button>
+            )}
+
+        </div>
```
polosanya, line 143:
```suggestion
            )}
        </div>
```

src/App.tsx
```diff
@@ -31,9 +31,33 @@ export function getReorderedGoods(
...
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+  // To avoid the original array mutation
+
   // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
```
Anastasiia-Svintsova, line 63: Remove the console.log statement as it is not needed.

src/App.tsx
```diff
@@ -42,54 +66,102 @@ export function getReorderedGoods(
...
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    const goods = getReorderedGoods(goodsFromServer, { sortType, isReversed });
+    const startOrder = sortType === SortType.NONE && isReversed === false;
```
Anastasiia-Svintsova, line 104: Use logical operators
```suggestion
    const startOrder = sortType === SortType.NONE && !isReversed;
```

src/App.tsx
```diff
@@ -42,54 +66,102 @@ export function getReorderedGoods(
...
+            {goods.map(good => (
+              <li
+                key={good}
+                data-cy="Good"
+              >
+                {good}
+              </li>
+            ))}
+          </ul>
         </ul>
```
Anastasiia-Svintsova, line 163: Hm, there is an `<ul>` inside another `<ul>`. 
Nested `<ul>` tags can lead to incorrect rendering and potential display issues in web browsers. Therefore, it is best to remove the nested `<ul>` and ensure that you have a proper HTML structure.

Anastasiia-Svintsova (changes requested):
Almost done! Check my comments, please

agniya-i (approved):

PR #1774, opened 2023-06-14
Solution

bohdan-mykhailenko:
[DEMO LINK](https://bohdan-mykhailenko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { PureComponent } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,70 +27,112 @@ type ReorderOptions = {
...
+              { 'is-light': sortType === SortType.ALPHABET })}
+          >
+            Sort alphabetically
+          </button>
+
+          <button
+            onClick={() => this.handleSortClick(SortType.LENGTH)}
+            type="button"
+            className={cn('button is-success',
+              { 'is-light': sortType === SortType.LENGTH })}
```
dinaminahi, line 101: It should be light when not selected and same for others

src/App.tsx
```diff
@@ -26,70 +27,112 @@ type ReorderOptions = {
...
+          <button
+            onClick={this.handleReverseClick}
+            type="button"
+            className={cn('button is-warning',
+              { 'is-light': isReversed })}
+          >
+            Reverse
+          </button>
+
+          {isDefaultOrder && (
```
dinaminahi, line 115: It is confusing, because we should show reset when it isn't default order
```suggestion
          {shouldShowReset && (
```
or
```suggestion
          {isOrderChanged && (
```

dinaminahi (changes requested):
Good job 👍 
But there are few things to fix

src/App.tsx
```diff
@@ -26,70 +27,112 @@ type ReorderOptions = {
...
+
+  handleReverseClick = () => {
+    this.setState(state => ({
+      isReversed: !state.isReversed,
+    }));
+  };
+
+  handleResetClick = () => {
+    this.setState({ sortType: SortType.NONE, isReversed: false });
+  };
```
naziamoff, line 74: no `handleClick` 🙅‍♀️ 🙅‍♀️ 🙅‍♀️ 

As a code reader I want to understand what happens in the logic of the function, but not where it's used

src/App.tsx
```diff
@@ -26,70 +27,112 @@ type ReorderOptions = {
...
+
+  render() {
+    const { sortType, isReversed } = this.state;
+    const shouldShowReset = isReversed || sortType !== SortType.NONE;
```
naziamoff, line 78: should show reset what?

naziamoff (changes requested):
good job, but let's fix naming a little

dinaminahi (approved):
Well done 👍

naziamoff (approved):

PR #1778, opened 2023-06-14
fix sortByLength

okosohor:
[DEMO LINK](https:/okosohor.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -21,75 +22,109 @@ enum SortType {
...
+      case SortType.LENGTH:
+        visibleGoods.sort((g1, g2) => (
+          g1.length - g2.length
+        ));
+
+        break;
+
+      default:
+        break;
+    }
```
naziamoff, line 74: let's create a separate method for that, it definitely should not be inside the render method

src/App.tsx
```diff
@@ -21,75 +22,109 @@ enum SortType {
...
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            onClick={this.sortByAlphabet}
+            type="button"
+            className={cn('button', 'is-info',
+              { 'is-light': sortType !== SortType.ALPHABET })}
+          >
```
naziamoff, line 88: no need in comma here
```suggestion
            className={cn('button is-info',
              { 'is-light': sortType !== SortType.ALPHABET })}
          >
```

naziamoff (approved):
good job 👍 
see some minor suggestions

PR #1785, opened 2023-06-14
solution

vitvadym:
[DEMO LINK](https://vitvadym.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,95 +1,99 @@
 import React from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { State } from './Type/State';
+import { getReorderedGoods } from './getReorderedGoods';
+import { SortType } from './Type/SortType';
+import { goodsFromServer } from './goodsFromServer';
 
-export const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
-];
+export class App extends React.Component<{}, State> {
+  state: State = {
+    isReversed: false,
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,95 +1,99 @@
...
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+          {(isReversed
+          || sortType !== SortType.NONE)
```
naziamoff, line 73: let's create a variable for that in the main logic

src/App.tsx
```diff
@@ -1,95 +1,99 @@
...
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {goods.map(good => (
+            <li
+              data-cy="Good"
+              key={good}
+            >
```
naziamoff, line 91: no need to move attributes to the new line here, you have only two of them and they will perfectly fit into a single line

src/getReorderedGoods.tsx
```diff
@@ -0,0 +1,31 @@
+import { ReorderOptions } from './Type/ReorderOptions';
+import { SortType } from './Type/SortType';
+
```
naziamoff, line 3: this file should have `.ts` extension, it's just a helper and has nothing to do with react

src/getReorderedGoods.tsx
```diff
@@ -0,0 +1,31 @@
...
+        return 0;
+    }
+  });
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  // eslint-disable-next-line no-console
+  console.log(sortType, isReversed);
```
naziamoff, line 28: do you need this log?

PR #1787, opened 2023-06-14
add solution

growingananas:
[DEMO LINK](https://growingananas.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -36,60 +37,131 @@ export function getReorderedGoods(
...
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={cn(
+              'button',
+              'is-info',
+              { 'is-light': sortType !== SortType.ALPHABET },
```
naziamoff, line 115: this way we have it readable and concise - I would stick to this formatting
```suggestion
            className={cn( 'button is-info', { 
              'is-light': sortType !== SortType.ALPHABET 
            },
```

src/App.tsx
```diff
@@ -36,60 +37,131 @@ export function getReorderedGoods(
...
+            onClick={this.sortByLength}
+          >
+            Sort by length
+          </button>
+
+          <button
+            type="button"
+            className={cn(
+              'button',
+              'is-warning',
```
naziamoff, line 138: see prev comment

src/App.tsx
```diff
@@ -36,60 +37,131 @@ export function getReorderedGoods(
...
+              { 'is-light': isReversed === false },
+            )}
+            onClick={this.reverse}
+          >
+            Reverse
+          </button>
+
+          {(isReversed === true || sortType !== SortType.NONE) && (
```
naziamoff, line 146: let's create a variable for that

naziamoff (approved):

src/App.tsx
growingananas: yeah..makes sense!

src/App.tsx
growingananas: const doReset = (isReversed === true || sortType !== SortType.NONE); 
{ doReset && (
            <button
              type="button"
              onClick={this.reset}
            >
              Reset
            </button>
          )}
        </div>

PR #1790, opened 2023-06-14
solution

kirilshorin:
[DEMO LINK](https://kirilshorin.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -34,62 +35,126 @@ export function getReorderedGoods(
...
+      isReversed: !state.isReversed,
+    }));
+  };
+
+  handleClickReset = () => (
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    })
+  );
```
naziamoff, line 87: let's think of better names for those variables.

- you definitely don't need to add `handle` for each method name if you perform only one action there
- you should avoid using `onSomething` name, as it does not give ANY info for the reader

src/App.tsx
```diff
@@ -34,62 +35,126 @@ export function getReorderedGoods(
...
+        <div className="buttons">
+          <button
+            type="button"
+            className={cn(
+              'button',
+              'is-info',
+              {
+                'is-light': sortType !== SortType.ALPHABET,
+              },
+            )}
```
naziamoff, line 105: you want it maximally concise
```suggestion
            className={cn('button is-info', {
               'is-light': sortType !== SortType.ALPHABET,
             })}
```

src/App.tsx
```diff
@@ -34,62 +35,126 @@ export function getReorderedGoods(
...
+          </button>
+
+          <button
+            type="button"
+            className={cn(
+              'button',
+              'is-success',
+              {
+                'is-light': sortType !== SortType.LENGTH,
+              },
```
naziamoff, line 118: see prev comment

src/App.tsx
```diff
@@ -34,62 +35,126 @@ export function getReorderedGoods(
...
+              {
+                'is-light': !isReversed,
+              },
+            )}
+            onClick={this.handleClickReverse}
+          >
+            Reverse
+          </button>
+
+          { isResetVisible && (
```
naziamoff, line 139:
```suggestion
          {isResetVisible && (
```

naziamoff (changes requested):
not bad, but let's polish

src/App.tsx
```diff
@@ -34,62 +35,114 @@ export function getReorderedGoods(
...
+      isReversed: !state.isReversed,
+    }));
+  };
+
+  getResetByClick = () => (
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    })
+  );
```
naziamoff, line 87: I don't think you are `getting` something in those functions. You sort, reset, or reverse. That's the action those method do - so it must be pretty obvious by their names. Also, is there any reason to add `byClick` in the end of the name?

naziamoff (changes requested):

33Orange (approved):
Well done

src/App.tsx
```diff
@@ -26,70 +27,119 @@ type ReorderOptions = {
   isReversed: boolean,
 };
 
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
```
roman-mirzoian, line 32: To specify the type of function arguments, it is better not to divide types into native (```goods: string[]```) and separately created (```{ sortType, isReversed }: ReorderOptions```), but to write a single type or interface for function arguments as a whole:
```
function getReorderedGoods(options: ReorderOptions) { ... }
```

src/App.tsx
```diff
@@ -26,70 +27,119 @@ type ReorderOptions = {
...
-        </button>
-
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+type State = {
```
roman-mirzoian, line 56: To set types of ```"sortType"``` and ```"isReversed"```, you already have a type ```"ReorderOptions"```. 
Better to name this type more generically and reuse it, rather than duplicating the typing with different names. (For example: ```SortingOptions```).

src/App.tsx
```diff
@@ -26,70 +27,119 @@ type ReorderOptions = {
...
+    const { isReversed, sortType } = this.state;
+    const isResetVisible = sortType !== SortType.NONE || isReversed;
+    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={cn('button is-info', {
```
roman-mirzoian, line 96: To improve the readability of the code, it is better to transfer the definition of classes to separate variables and use the ready-made result. You can see an example of use on the website of the [classnames ](https://www.npmjs.com/package/classnames) package.
This applies to all code.

src/App.tsx
```diff
@@ -26,70 +27,119 @@ type ReorderOptions = {
...
+              'is-light': sortType !== SortType.ALPHABET,
+            })}
+            onClick={() => this.sort(SortType.ALPHABET)}
```
roman-mirzoian, line 99: To improve code readability, it is better to move the onClick functions of handlers to separate handlers.
This applies to all code.

roman-mirzoian (changes requested):
The component is nicely updated, but some updates can be made to optimize the code.

naziamoff (approved):

PR #1796, opened 2023-06-15
add task solution

KatyaZaremska:
[DEMO LINK](https://KatyaZaremska.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
@@ -26,7 +27,6 @@ type ReorderOptions = {
   isReversed: boolean,
 };
 
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,70 +27,117 @@ type ReorderOptions = {
...
+}
+
+export class App extends React.PureComponent<{}, State> {
+  state: Readonly<State> = {
+    sortType: SortType.NONE,
+    isReversed: false,
+  };
+
+  handleSortByAlphabet = () => {
+    this.setState({ sortType: SortType.ALPHABET });
```
naziamoff, line 69: I don't think there's a reason to add `handle` before the name of each method
```suggestion
  sortByAlphabet = () => {
    this.setState({ sortType: SortType.ALPHABET });
```

src/App.tsx
```diff
@@ -26,70 +27,117 @@ type ReorderOptions = {
...
+  };
+
+  handleSortByLenght = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  handleReverse = () => {
+    this.setState(state => ({ isReversed: !state.isReversed }));
+  };
```
naziamoff, line 78: same here. You just set state here, it does not do any additional operations. So no need in `handle`

src/App.tsx
```diff
@@ -26,70 +27,117 @@ type ReorderOptions = {
...
+    this.setState({
+      sortType: SortType.NONE,
+      isReversed: false,
+    });
+  };
+
+  render() {
+    const { sortType, isReversed } = this.state;
+    const goods = getReorderedGoods(goodsFromServer, this.state);
+    const isVisibleReset = isReversed || sortType !== SortType.NONE;
```
naziamoff, line 90:
```suggestion
    const isResetVisible = isReversed || sortType !== SortType.NONE;
```

src/App.tsx
```diff
@@ -26,70 +27,117 @@ type ReorderOptions = {
...
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={classNames('button is-info', {
+              'is-light': sortType !== SortType.ALPHABET,
+            })}
```
naziamoff, line 99: 🔥

naziamoff (approved):
good job 👍 
See minor suggestions

PR #1807, opened 2023-06-15
solution

thienlaoo:
[DEMO LINK](https://thienlaoo.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -47,49 +57,89 @@ export function getReorderedGoods(
...
+  sortByAlphabet = () => {
+    this.setState({ sortType: SortType.ALPHABET });
+  };
+
+  sortByLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  reverseGoods = () => {
+    this.setState({ isReversed: true });
```
dinaminahi, line 75: image

src/App.tsx
```diff
@@ -47,49 +57,89 @@ export function getReorderedGoods(
...
+              type="button"
+              className={cn('button is-warning', {
+                'is-light': !isReversed,
+              })}
+              onClick={this.reverseGoods}
+            >
+              Reverse
+            </button>
+
+            {(sortType !== SortType.NONE || isReversed) && (
```
dinaminahi, line 122: Can we move this into a variable so it will be more readable?

dinaminahi (changes requested):
Good job 👍 
There are few things we need to fix

PR #1808, opened 2023-06-15
added task solution

holytimy4:
[DEMO LINK](https://holytimy4.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -21,75 +22,119 @@ enum SortType {
...
+          <button
+            type="button"
+            onClick={this.changeOrder}
+            className={cn('button is-warning',
+              { 'is-light': !isReversed })}
+          >
+            Reverse
+          </button>
+
+          {(isReversed || sortType !== SortType.NONE) && (
```
33Orange, line 117: Let's move this condition to a variable and use it here

33Orange (changes requested):
Demo link doesn't work

image

PR #1809, opened 2023-06-16
preload

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
@@ -21,75 +22,88 @@ enum SortType {
   LENGTH,
 }
 
-type ReorderOptions = {
-  sortType: SortType,
-  isReversed: boolean,
-};
+export class App extends React.Component {
+  goods = [...goodsFromServer];
 
-// Use this function in the render method to prepare goods
-export function getReorderedGoods(
-  goods: string[],
-  { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
+  state = {
+    isReversed: false,
+    sortType: 0,
+  };
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  buttonClickReverse = (): void => {
+    this.goods = this.goods.reverse();
+    this.setState(prevState => (
+      { ...prevState, isReversed: !prevState.isReversed }));
+  };
 
-  return visibleGoods;
-}
+  buttonClickAbc = (): void => {
+    this.goods = [...this.goods.sort((a, b) => a.localeCompare(b))];
+    this.setState(prevState => ({ ...prevState, sortType: SortType.ALPHABET }));
+  };
 
-// DON'T save goods to the state
-// type State = {
-//   isReversed: boolean,
-//   sortType: SortType,
-// };
+  buttonClickLength = (): void => {
+    this.goods = [...this.goods.sort((a, b) => a.length - b.length)];
+    this.setState(prevState => ({ ...prevState, sortType: SortType.LENGTH }));
+  };
 
-export const App: React.FC = () => {
-  return (
-    <div className="section content">
-      <div className="buttons">
-        <button
-          type="button"
-          className="button is-info is-light"
-        >
-          Sort alphabetically
-        </button>
+  buttonClickReset = (): void => {
+    this.goods = [...goodsFromServer];
+    this.setState(prevState => (
+      { ...prevState, isReversed: false, sortType: SortType.NONE }));
+  };
 
-        <button
-          type="button"
-          className="button is-success is-light"
-        >
-          Sort by length
-        </button>
+  render() {
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            id="SortByAbc"
+            className={`button is-info ${(this.state.sortType === 1 ? '' : 'is-light')}`}
+            onClick={this.buttonClickAbc}
+          >
+            Sort alphabetically
+          </button>
 
-        <button
-          type="button"
-          className="button is-warning is-light"
-        >
-          Reverse
-        </button>
+          <button
+            type="button"
+            id="SortByLength"
+            className={`button is-success ${(this.state.sortType === 2 ? '' : 'is-light')}`}
+            onClick={this.buttonClickLength}
+          >
+            Sort by length
+          </button>
 
-        <button
-          type="button"
-          className="button is-danger is-light"
-        >
-          Reset
-        </button>
-      </div>
+          <button
+            type="button"
+            id="reverse"
+            onClick={this.buttonClickReverse}
+            className={`button is-warning ${(this.state.isReversed ? '' : 'is-light')}`}
+          >
+            Reverse
+          </button>
 
-      <ul>
+          {(this.state.isReversed || this.state.sortType !== 0)
+          && (
+            <button
+              type="button"
+              id="reset"
+              className="button is-danger is-light"
+              onClick={this.buttonClickReset}
+            >
+              Reset
+            </button>
+          )}
+        </div>
+
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {this.goods.map(productName => (
+              <li data-cy="Good" key={productName}>{productName}</li>
+            ))}
+          </ul>
         </ul>
-      </ul>
-    </div>
-  );
-};
+      </div>
+    );
+  }
+}
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Implement the ability to reorder the goods with the given buttons.
     - remove `is-light` class from the button when this order is applied;
 1. `Sort by length` button should sort the goods by length.
     - remove `is-light` class from the button when this order is applied;
-1. `Reverse` button changes the current order to the opposite one. 
+1. `Reverse` button changes the current order to the opposite one.
     - it should work together with any sorting;
     - remove `is-light` class from the button when goods are reversed;
     - the second click should show goods in the direct order;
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -21,75 +22,96 @@ enum SortType {
...
+  sortType: number,
 };
 
-// Use this function in the render method to prepare goods
-export function getReorderedGoods(
-  goods: string[],
-  { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
```
sTorba24, line 35: you should use function to prepare goods

src/App.tsx
```diff
@@ -21,75 +22,96 @@ enum SortType {
...
+    this.setState(prevState => (
+      { ...prevState, isReversed: !prevState.isReversed }));
+  };
 
-// DON'T save goods to the state
-// type State = {
-//   isReversed: boolean,
-//   sortType: SortType,
-// };
+  buttonClickAbc = (): void => {
```
sTorba24, line 44: use more meaningful name for function

src/App.tsx
```diff
@@ -21,75 +22,96 @@ enum SortType {
...
-  return (
-    <div className="section content">
-      <div className="buttons">
-        <button
-          type="button"
-          className="button is-info is-light"
-        >
-          Sort alphabetically
-        </button>
+  buttonClickLength = (): void => {
```
sTorba24, line 49:
```suggestion
  handleSortByLength = (): void => {
```

src/App.tsx
```diff
@@ -21,75 +22,96 @@ enum SortType {
...
-        </button>
+  render() {
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            id="SortByAbc"
+            className={`button is-info
+              ${(this.state.sortType === SortType.ALPHABET ? '' : 'is-light')}`}
```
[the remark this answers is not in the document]
victor-buldenko: what for? same code without useless module.

src/App.tsx
```diff
@@ -15,81 +16,132 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
-      </div>
+type State = {
+  isReversed: boolean,
+  sortType: number,
+};
+
+export class App extends React.PureComponent<{}, State> {
+  goods = [...goodsFromServer];
```
sTorba24, line 63: what a sense to keep it here?

src/App.tsx
```diff
@@ -15,81 +16,132 @@ export const goodsFromServer = [
...
+    this.setState(prevState => ({ ...prevState, sortType: SortType.LENGTH }));
+  };
+
+  buttonClickResetGoodsList = (): void => {
+    this.setState(prevState => (
+      { ...prevState, isReversed: false, sortType: SortType.NONE }));
+  };
+
+  render() {
+    this.goods = getReorderedGoods(this.goods, this.state);
```
sTorba24, line 89:
```suggestion
    const goods = getReorderedGoods(goodsFromServer, this.state);
```

src/App.tsx
```diff
@@ -15,81 +16,132 @@ export const goodsFromServer = [
...
+
+export class App extends React.PureComponent<{}, State> {
+  goods = [...goodsFromServer];
+
+  state = {
+    isReversed: false,
+    sortType: 0,
+  };
+
+  buttonClickReverse = (): void => {
```
sTorba24, line 70:
```suggestion
  reverse = (): void => {
```

src/App.tsx
```diff
@@ -15,81 +16,132 @@ export const goodsFromServer = [
...
+    this.setState(prevState => (
+      { ...prevState, isReversed: !prevState.isReversed }));
+  };
+
+  buttonClickByAlphabet = (): void => {
```
sTorba24, line 75:
```suggestion
  sortByAlphabet = (): void => {
```

src/App.tsx
```diff
@@ -15,81 +16,132 @@ export const goodsFromServer = [
...
+    this.setState(prevState => ({ ...prevState, sortType: SortType.ALPHABET }));
```
sTorba24, line 76: why do you need prevState ?

src/App.tsx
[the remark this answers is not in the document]
sTorba24: because you will not write like if .... - add '' , what a reason to do so?
You can have more conditions in future and sense to use classnames  will increase

sTorba24 (changes requested):
Fix my comments. Not needed to resolve conversation by yourself, one of mentors'll do it

src/App.tsx
```diff
@@ -15,81 +16,132 @@ export const goodsFromServer = [
...
-
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
-) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
+): string[] {
+  let visibleGoods = [...goods];
```
sTorba24, line 34:
```suggestion
  const visibleGoods = [...goods];
```

src/App.tsx
```diff
@@ -15,81 +16,132 @@ export const goodsFromServer = [
...
+
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+    default:
+      visibleGoods = [...goodsFromServer];
```
sTorba24, line 46: it's already goodsFromServer. Don't take variable outside this function

sTorba24 (changes requested):

src/App.tsx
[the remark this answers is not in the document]
victor-buldenko: when I obtain more than 1 extra class in the future - then I use react classname module, and now it's not needed.

src/App.tsx
victor-buldenko: reverse what? display? array? table? world?
image

src/App.tsx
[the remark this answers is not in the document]
sTorba24: it's good practice train yourself to use the right tools right away

src/App.tsx
victor-buldenko: You can have more conditions in future and this make sense.

src/App.tsx
sTorba24: it's not array reversing, so use just it.
You can use any name you prefer, but it's not click

src/App.tsx
victor-buldenko: image

src/App.tsx
victor-buldenko: i don`t have another way

src/App.tsx
sTorba24: you're reassigning the same value, i see

src/App.tsx
sTorba24: You're not using prevState, why do you need functional setState? This task hasn't future

sTorba24 (changes requested):

src/App.tsx
sTorba24: you can just remove it

victor-buldenko:
This task hasn't future oh really? So why do you tell me about the future when speak about class names?
image
... what about logical sequence?

src/App.tsx
victor-buldenko: Do you have another way? ... show me.

sTorba24 (changes requested):
You have at least 1 case where classNames can be used, so it make sense here. More precisely, there are 3 of them.
And 0 usages of prevState. Don't think there is a reason to continue it.
Do it right and  it'll be approved

src/App.tsx
sTorba24: image
still not resolved

src/App.tsx
sTorba24: what do you mean?

src/App.tsx
```diff
@@ -15,81 +16,131 @@ export const goodsFromServer = [
...
+  buttonClickReverse = (): void => {
+    this.setState(prevState => (
+      { isReversed: !prevState.isReversed }));
+  };
+
+  sortByAlphabet = (): void => {
+    this.setState(prevState => ({ ...prevState, sortType: SortType.ALPHABET }));
+  };
+
+  buttonClickByLength = (): void => {
```
sTorba24, line 79: it's not click. It's handler. Fix everywhere

src/App.tsx
```diff
@@ -15,81 +16,131 @@ export const goodsFromServer = [
...
+    this.goods = getReorderedGoods(goodsFromServer, this.state);
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            id="SortByAbc"
+            className={`button is-info
+              ${(this.state.sortType === SortType.ALPHABET ? '' : 'is-light')}`}
```
sTorba24, line 97: destructure state.

src/App.tsx
```diff
@@ -15,81 +16,131 @@ export const goodsFromServer = [
...
+  buttonClickByLength = (): void => {
+    this.setState(prevState => ({ ...prevState, sortType: SortType.LENGTH }));
+  };
+
+  buttonClickResetGoodsList = (): void => {
+    this.setState({ isReversed: false, sortType: SortType.NONE });
+  };
+
+  render() {
+    this.goods = getReorderedGoods(goodsFromServer, this.state);
```
sTorba24, line 88:
```suggestion
    const goods = getReorderedGoods(goodsFromServer, this.state);
```

src/App.tsx
```diff
@@ -15,81 +16,131 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
-      </div>
+type State = {
+  isReversed: boolean,
+  sortType: number,
+};
+
+export class App extends React.PureComponent<{}, State> {
+  goods: string[] = [];
```
sTorba24, line 63:
```suggestion
```

sTorba24 (changes requested):

src/App.tsx
victor-buldenko: What about RESET?

src/App.tsx
victor-buldenko: Hmmm... how reset works in this case?

src/App.tsx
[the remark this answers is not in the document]
victor-buldenko: I have enough practice, save this for yourself.

src/App.tsx
sTorba24: still not resolved

sTorba24 (approved):
Amazing! 🔥 🔥 🔥

PR #1810, opened 2023-06-18
add solution

allagutsul27:
[DEMO LINK](https://allagutsul27.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -31,65 +32,126 @@ export function getReorderedGoods(
...
+      case ALPHABET:
+        return good1.localeCompare(good2);
+
+      case LENGTH:
+        return good1.length - good2.length;
+
+      default:
+        return 0;
+    }
+  });
```
IvanFesenko, line 50: No need tto sort if `sortType` is NONE

IvanFesenko (changes requested):

PR #1821, opened 2023-06-26
added solution

sheva10barca:
[DEMO LINK](https://sheva10barca.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -42,6 +42,7 @@ export function getReorderedGoods(
       case SortType.LENGTH:
         return good1.length - good2.length;
 
+      case SortType.NONE:
       default:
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -22,74 +24,126 @@ enum SortType {
...
+      case SortType.ALPHABET:
+        return good1.localeCompare(good2);
+
+      case SortType.LENGTH:
+        return good1.length - good2.length;
+
+      default:
+        return 0;
+    }
+  });
```
IvanFesenko, line 48: No need to sort in case SortType is NONE

IvanFesenko (changes requested):

PR #1840, opened 2023-07-24
Add solution

ArtemBehter:


src/App.tsx
```diff
@@ -26,70 +26,121 @@ type ReorderOptions = {
...
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
 
   return visibleGoods;
 }
 
-// DON'T save goods to the state
```
DorotaLeniecDev, line 44: Refactor your solution and follow this rule `// DON'T save goods to the state`

DorotaLeniecDev (changes requested):
Check my comment, refacotr following the task rule :)

src/App.tsx
```diff
@@ -26,70 +26,120 @@ type ReorderOptions = {
...
   goods: string[],
   { sortType, isReversed }: ReorderOptions,
 ) {
-  // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  if (sortType === SortType.ALPHABET) {
```
darokrk, line 35: maybe instead of using so many if's better will be used switch the check statement? You are using sort() method in every scenario but with different content:

```
visibleGoods.sort((a, b) => {
    switch (sortType) {
      case SortType.ALPHABET:
        return a.localeCompare(b);
      case SortType.LENGTH:
        return a.length - b.length;
      default:
        return SortType.NONE;
    }
  });
```

src/App.tsx
```diff
@@ -26,70 +26,120 @@ type ReorderOptions = {
...
+    visibleGoods.sort((a, b) => (a.localeCompare(b)));
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => {
+      if (a.length - b.length === 0) {
```
darokrk, line 41: think this check `b.length === 0` is redundant here

src/App.tsx
```diff
@@ -26,70 +26,120 @@ type ReorderOptions = {
...
+export class App extends React.Component<{}, State> {
+  state: State = {
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
+
+  sortAlphabetically = () => {
+    this.setState(() => ({
+      sortType: SortType.ALPHABET,
+      isReversed: false,
```
darokrk, line 70: think you don't need to define this reverse statement is this sort functions, only in reverseGoods(), so it can be removed :)

src/App.tsx
```diff
@@ -26,70 +26,120 @@ type ReorderOptions = {
...
+    this.setState(() => ({
+      sortType: SortType.NONE,
+      isReversed: false,
+    }));
+  };
+
+  sortByLength = () => {
+    this.setState(() => ({
+      sortType: SortType.LENGTH,
+      isReversed: false,
```
darokrk, line 84: this isReversed: false, can be removed here

darokrk (changes requested):
UI looks good 👍  but code needs some improvements 👨🏽‍💻 🛠️

darokrk (approved):
Good job 👏🏽  🔥

PR #1841, opened 2023-08-09
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-access-state-in-setstate */
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
@@ -16,9 +17,9 @@ export const goodsFromServer = [
 ];
 
 enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
+  NONE = 'none',
+  ALPHABET = 'alphabet',
+  LENGTH = 'length',
 }
 
 type ReorderOptions = {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,70 +27,122 @@ type ReorderOptions = {
...
+  sortAlphabet = () => {
+    this.setState({ sortType: SortType.ALPHABET });
+  };
+
+  sortLength = () => {
+    this.setState({ sortType: SortType.LENGTH });
+  };
+
+  reverse = () => {
+    this.setState({ isReversed: !this.state.isReversed });
```
darokrk, line 66: Avoid directly accessing this.state in the setState function, as state updates may be asynchronous. Use a function instead.

```
 this.setState(state => ({
    }));
```

src/App.tsx
```diff
@@ -26,70 +27,122 @@ type ReorderOptions = {
...
+  };
+
+  reset = () => {
+    this.setState({ isReversed: false });
```
darokrk, line 70: You are calling 'setState' twice which might lead to unnecessary re-renders and performance issue. You should combine these two into one.

```
reset = () => {
    this.setState({
      sortType: SortType.NONE,
      isReversed: false,
    });
  };
```

src/App.tsx
```diff
@@ -26,70 +27,122 @@ type ReorderOptions = {
...
+    this.setState({ sortType: SortType.NONE });
+  };
+
+  render() {
+    const { isReversed, sortType } = this.state;
+    let visibleGoods = [...goodsFromServer];
+
+    if (sortType === SortType.ALPHABET) {
```
darokrk, line 78: You are currently recreating the 'visibleGoods' array and then potentially overriding it later. Consider creating it only once for performance reason. Don't need to create that `if else` statements, this logic will happen in `getReorderedGoods `function directly in switch statement, just try here and base on current state:

`const visibleGoods = getReorderedGoods(goodsFromServer, this.state);`

src/App.tsx
```diff
@@ -26,70 +27,122 @@ type ReorderOptions = {
...
-  console.log(sortType, isReversed);
+  getReorderedGoods = (
+    goods: string[],
+    { sortType, isReversed }: ReorderOptions,
+  ) => {
+    const visibleGoods = [...goods];
 
-  return visibleGoods;
-}
+    if (sortType === SortType.ALPHABET) {
```
darokrk, line 42: you can simplify this logic by using `switch` statement:

```
  visibleGoods.sort((a, b) => {
    switch (sortType) {
      case SortType.ALPHABET:
        return a.localeCompare(b);
      case SortType.LENGTH:
        return a.length - b.length;
      default:
        return SortType.NONE;
    }
  });

  if (isReversed) {
    visibleGoods.reverse();
  }

  return visibleGoods;
};
```

src/App.tsx
```diff
@@ -26,70 +27,122 @@ type ReorderOptions = {
...
+
+    if (sortType === SortType.ALPHABET) {
+      visibleGoods = this.getReorderedGoods(goodsFromServer,
+        { sortType: SortType.ALPHABET, isReversed });
+    } else if (sortType === SortType.LENGTH) {
+      visibleGoods = this.getReorderedGoods(goodsFromServer,
+        { sortType: SortType.LENGTH, isReversed });
+    }
+
+    if (isReversed) {
```
darokrk, line 86: this is redundant, please remove it to not duplicate logic :)

src/App.tsx
```diff
@@ -26,70 +27,122 @@ type ReorderOptions = {
...
+            className={isReversed === true
+              ? 'button is-warning'
+              : 'button is-warning is-light'}
+            onClick={this.reverse}
+          >
+            Reverse
+          </button>
+
+          {(isReversed || sortType !== SortType.NONE) && (
+            <>
```
darokrk, line 125: think this empty brackets are redundant here

src/App.tsx
```diff
@@ -26,70 +27,122 @@ type ReorderOptions = {
...
 
-      <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
```
darokrk, line 139: here we have now `<ul>` inside `<ul>` think one is enough so this can be removed :)

darokrk (changes requested):
Not bad 😃  few improvements needed and we are good to go 💪🏽 🛠️ !

src/App.tsx
```diff
@@ -39,13 +39,28 @@ export class App extends React.PureComponent {
...
     const visibleGoods = [...goods];
 
-    if (sortType === SortType.ALPHABET) {
-      visibleGoods.sort();
-    }
-
-    if (sortType === SortType.LENGTH) {
-      visibleGoods.sort((a, b) => a.length - b.length);
-    }
+    // if (sortType === SortType.ALPHABET) {
```
darokrk, line 42: please remove all commented code :)

darokrk (changes requested):
Nice ! 👍🏽  please remove all commented old code and we got this 🏆  😎

darokrk (approved):
Good job 💪🏽

PR #1843, opened 2023-08-15
Add sorts

AmadeuszAndroid:
[DEMO LINK](https://amadeuszandroid.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { GoodsList } from './components/GoodsList';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -34,62 +35,102 @@ export function getReorderedGoods(
...
+  reset = () => {
+    this.setState({
+      isReversed: false,
+      sortType: SortType.NONE,
+    });
+  };
+
+  render() {
+    const { sortType, isReversed } = this.state;
+    const isLight = 'is-light';
```
Radoslaw-Czerniawski, line 88: Don't use variables for simple conditions, declare those on the props

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,13 @@
+import { Good } from './Good';
+
+type GoddsListProps = {
+  goods: string[]
+};
+
+export const GoodsList = ({ goods }: GoddsListProps) => {
+  return (
+    <ul>
+      {goods.map(good => <Good name={good} key={goods.indexOf(good)} />)}
```
Radoslaw-Czerniawski, line 10: [.map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) as the second argument receives index of the currently iterated element, so there is no need for the indexOf
```tsx
{goods.map((good, index) => (
  <Good name={good} key={index} />
))}
```
however you should NEVER use indexes as keys, goods in this case are unique so use the values as keys (also do the implicit return)
```tsx
export const GoodsList = ({ goods }: GoddsListProps) => (
  <ul>
    {goods.map((good) => (
      <Good name={good} key={good} />
    ))}
  </ul>
);
```

Radoslaw-Czerniawski (changes requested):
Works great :D, some code changes needed though

src/App.tsx
```diff
@@ -34,62 +35,102 @@ export function getReorderedGoods(
...
+  sortByLength = () => {
+    this.setState({
+      sortType: SortType.LENGTH,
+    });
+  };
+
+  reverse = () => {
+    const previousIsRevesved = this.state.isReversed;
+
+    this.setState({
```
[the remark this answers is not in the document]
AmadeuszAndroid: I edited state in function. `...prevState` is redundant.

src/components/GoodsList.tsx
AmadeuszAndroid: I deleted return and switched keys to goods.

src/App.tsx
AmadeuszAndroid: I changed to `classNames`.

src/App.tsx
```diff
@@ -34,62 +37,99 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  if (sortType === SortType.ALPHABET) {
+    visibleGoods.sort((g1, g2) => g1.localeCompare(g2));
+  } else if (sortType === SortType.LENGTH) {
```
carlos-cne, line 42: when we use the `else if` statement implies to have an `else` statement on the end. Like:
```js
if (condition) {
  // doSomething()
} else if (otherCondition) {
  // doOtherThing()
} else { // if any condition matches
  // doThis()
}

File: src/App.tsx [hunk 5]
```diff
@@ -34,62 +37,99 @@ export function getReorderedGoods(
...
+    this.setState({
+      sortType: SortType.LENGTH,
+    });
+  };
+
+  reverse = () => this.setState((prevState: State) => ({
+    isReversed: !prevState.isReversed,
+  }));
+
+  reset = () => {
```
carlos-cne, line 77: to avoid duplication, you could set an initialState constant and use it as your initial state and on the reset method:
```ts
const INITIAL_STATE: State = {
  ...
}
state: Readonly<State> = INITAL_STATE;
reset = () => this.setState(INITIAL_STATE);
```

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,11 @@
+import { Good } from './Good';
+
+type GoddsListProps = {
+  goods: string[]
+};
+
+export const GoodsList = ({ goods }: GoddsListProps) => (
+  <ul>
```
carlos-cne, line 8: You are using semantic HTML. Great job!
Using `ul` and `li` to represent a list of items will help your code and also improve the user experience because the browser will understand better the elements on the page and it also improves the page accessibility.

carlos-cne (approved):
Your last changes look great and it works as expected. 
Great job!

src/App.tsx
AmadeuszAndroid: We have only 2 kinds of sort. What logic do You expect on `else`?

PR #1844, opened 2023-08-19
develop

domkab:
https://domkab.github.io/react_list-of-goods

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -26,69 +26,92 @@ type ReorderOptions = {
...
-          Reset
-        </button>
+        {sortType !== SortType.NONE && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setIsReversed(false);
+              setSortType(SortType.NONE);
+            }}
```
Anastasiia-Svintsova, line 104: Move the onClick logic to a separate function. It follows the best practice of separating concerns and keeping your JSX code clean and focused on rendering components rather than defining complex logic directly within it.

Anastasiia-Svintsova (changes requested):
- You should pass this test. The reset button should be visible if the reverse button is active 
image
image

Anastasiia-Svintsova (approved):

PR #1849, opened 2023-09-04
add task solution

ogospadzipalina:
[DEMO LINK](https://ogospadzipalina.github.io/react_list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -108,7 +108,7 @@ export class App extends React.Component<{}, State> {
             type="button"
             className={this.state.sortType === SortType.LENGTH
               ? 'button is-info'
-              : 'button is-info is-success'}
+              : 'button is-info is-success is-light'}
             onClick={this.sortByLength}
           >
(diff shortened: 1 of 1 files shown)


Dolnys (changes requested):
It looks good, but you have the wrong classes for the sorting button by length and reverse. It should be "is-success" for sorting by length and "is-warning" for reverse sorting.

Dolnys:
It's better, but you are still missing styles for the selected buttons. The 'is-info' class should only be used with the 'sorting by alphabet' button. You should remove 'is-info' from here:` 'button is-info is-success is-light'` and here: `'button is-info is-warning is-light'`

choeqq (changes requested):
Make sure to apply changes suggested above, and you're good to go⬆️

ogospadzipalina:
I got lost in the styles somehow, thank you for comments!

choeqq (approved):

PR #1851, opened 2023-09-04
Solution

Sancz0pansa:
[DEMO LINK](https://Sancz0pansa.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


ihor-jpeg (changes requested):
test

ogospadzipalina (approved):
looks good to me

src/App.tsx
```diff
@@ -26,70 +27,107 @@ type ReorderOptions = {
...
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          <ul>
+            {getReorderedGoods(goodsFromServer,
+              { sortType, isReversed }).map(good => (
+            // eslint-disable-next-line @typescript-eslint/indent
```
pawel-peksa, line 125: I think you can find a solution without disabling eslint here

src/App.tsx
```diff
@@ -26,70 +27,107 @@ type ReorderOptions = {
...
+          <button
+            type="button"
+            className={classNames('button',
+              'is-info', { 'is-light': isReversed === false })}
+            onClick={() => this.setReverse()}
+          >
+            Reverse
+          </button>
+
+          {(sortType !== 0 || isReversed) && (
```
pawel-peksa, line 107: Nitpick: When the logic for rendering consist of multiple logical operators, you could lift it above return statement. e.g.
`const isResetVisible = (sortType !== 0 || isReversed)`
Then there is less clutter in JSX

src/App.tsx
Sancz0pansa: when I add spaces, it asks me to remove them and so on

choeqq (approved):

PR #1855, opened 2023-09-05
task solution

HubertKlamann:
[DEMO LINK](https://HubertKlamann.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -31,34 +31,27 @@ type ReorderOptions = {
 export function getReorderedGoods(
   { sortType, isReversed }: ReorderOptions,
 ) {
-  let visibleGoods = [...goodsFromServer];
+  const visibleGoods = [...goodsFromServer];
 
-  if (sortType === SortType.ALPHABET) {
-    visibleGoods = visibleGoods.sort((a, b) => a.localeCompare(b));
-  } else if (sortType === SortType.LENGTH) {
-    visibleGoods = visibleGoods.sort((a, b) => {
-      const lengthDiffrence = a.length - b.length;
+  visibleGoods.sort((a, b) => {
+    if (sortType === SortType.ALPHABET) {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -28,11 +29,27 @@ type ReorderOptions = {
...
+    visibleGoods = visibleGoods.sort((a, b) => {
+      const lengthDiffrence = a.length - b.length;
+
+      if (lengthDiffrence === 0) {
+        return a.localeCompare(b);
+      }
+
+      return lengthDiffrence;
+    });
+  }
```
Matylda-Lewandowska, line 48: You don't need to use `let visibleGoods` and then later `visibleGoods = visibleGoods.sort(...)`, with `const visibleGoods` should work just with ` visibleGoods.sort(...)`, because `.sort()`, `reverse()`, and couple more operations are mutating the original array, here:  `const visibleGoods`  and will change how the array looks :)

Matylda-Lewandowska (approved):
All over, I think code looks good to me :)

src/App.tsx
asakevych, line 48: 1. I agree with Matylda, it can be const and then you shouldn't assign value to it, just use the method to modify it
2. you can simplify your function using ifs inside sort method

```suggestion
  const visibleGoods = [...goodsFromServer];

  visibleGoods.sort((goodA, goodB) => {
    switch (sortType) {
      case SortType.ALPHABET:
        return goodA.localeCompare(goodB);
      case SortType.LENGTH:
        return goodA.length - goodB.length;
      default:
        return 0;
    }
  });

  if (isReversed) {
    visibleGoods.reverse();
  }
```

asakevych (changes requested):
Good job!! 🚀 just some small changes can be provided 😉

PR #1857, opened 2023-09-05
develop

Matylda-Lewandowska:
[DEMO LINK](https://matylda-lewandowska.github.io/react_list-of-goods/)

task solution - list of goods

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Implement the ability to reorder the goods with the given buttons.
     - remove `is-light` class from the button when this order is applied;
 1. `Sort by length` button should sort the goods by length.
     - remove `is-light` class from the button when this order is applied;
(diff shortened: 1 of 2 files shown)


Vira-V (approved):
Everything looks great to me.
I would recommend changing one line in all buttons. You used an arrow function as the event handler. When the element with this onClick attribute is clicked, it will execute the arrow function, which in turn calls the` this.sortReset() method`. This approach creates a new function every time the component renders.       In this case, I guess is better to use this form `onClick={this.sortReset}`.  When the element is clicked, it will call the sortReverse method directly.

choeqq (approved):
Please take a look at comment above ⬆️ @Matylda-Lewandowska

Matylda-Lewandowska:
I've made these adjustments. :)

PR #1863, opened 2023-09-06
solution

Capoverde:
[DEMO](https://Capoverde.github.io/react_list-of-goods/)


choeqq (changes requested):
That solution is attached to wrong homework task. we want to be able to sort and reset our list, as said in readme -> https://github.com/mate-academy/react_goods-selector

asakevych (approved):
Great job!! 🚀 🔥 👏

PR #1871, opened 2023-09-08
add task solution

arturwisnia:
[demo](https://arturwisnia.github.io/react_list-of-goods/)


asakevych (approved):
Great job!! 🚀 🔥 👏 
please remember to give descriptive names to variables (when mapping through goods its elements coul be `good` instead of `v`)

PR #1885, opened 2023-12-06
Solution

WiktoriaBielec:
[DEMO LINK](https://WiktoriaBielec.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -82,9 +82,9 @@ export class App extends React.Component<{}, State> {
 
   handleReset = () => {
     this.setState({
-      sortType: SortType.NONE, 
+      sortType: SortType.NONE,
       isReversed: false,
-    })
+    });
   };
 
   render() {


src/App.tsx
```diff
@@ -31,65 +31,104 @@ export function getReorderedGoods(
...
+    visibleGoods = visibleGoods.sort((a, b) => {
+      return b.localeCompare(a);
+    });
+  } else if (sortType === 2 && isReversed === true) {
+    visibleGoods = visibleGoods.sort((a, b) => {
+      return a.length - b.length;
+    }).reverse();
+  } else if (sortType === 0 && isReversed === true) {
+    visibleGoods = visibleGoods.reverse();
+  }
```
MarcinLigmann, line 54: you can also use a switch case statement and finally use an if statement for isReversed
something like that:
``` switch (sortType) {
    case SortType.ALPHABET:
      break;
    case ...
  }

  if (isReversed) {
    visibleGoods.reverse();
  }

File: src/App.tsx [hunk 2]
```diff
@@ -31,65 +31,104 @@ export function getReorderedGoods(
...
+    const goods = getReorderedGoods(goodsFromServer, this.state);
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={this.state.sortType === 1
+              ? 'button is-info'
+              : 'button is-info is-light'}
```
MarcinLigmann, line 83: You can use the && operator however, this solution is equally good
I think that using the values from the SortType enum will be a better solution than 0, 1 ,2...
```className={`button is-info ${(sortType !== SortType.ALPHABET) && ('is-light')}`}```

File: src/App.tsx [hunk 3]
```diff
@@ -31,65 +31,104 @@ export function getReorderedGoods(
...
+          </button>
+
+          <button
+            type="button"
+            className={this.state.isReversed === true
+              ? 'button is-warning'
+              : 'button is-warning is-light'}
+            onClick={() => (this.state.isReversed
+              ? this.setState({ isReversed: false })
+              : this.setState({ isReversed: true }))}
```
MarcinLigmann, line 106: I believe that such things should be stored in the functions described before render()

MarcinLigmann (approved):
These are just my tips that I used in my code.
Otherwise the code works :)

src/App.tsx
WiktoriaBielec: I think my solution is clearer than &&

src/App.tsx
WiktoriaBielec: 'if' was used as the example at the current topic on mate. Also I'm using && at almost all of them so using 'case' could be confusing.

src/App.tsx
```diff
@@ -31,65 +31,104 @@ export function getReorderedGoods(
...
+
+  render() {
+    const goods = getReorderedGoods(goodsFromServer, this.state);
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={this.state.sortType === 1
```
mbruhler, line 81: use enums, not numbers

src/App.tsx
```diff
@@ -31,65 +31,115 @@ export function getReorderedGoods(
...
+
+          {(this.state.sortType !== SortType.NONE
+          || this.state.isReversed === true)
+        && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => this.setState({
+              sortType: SortType.NONE, isReversed: false,
+            })}
```
mvjl000, line 127: Create a handler function for this as well ;)

src/App.tsx
```diff
@@ -31,65 +31,115 @@ export function getReorderedGoods(
...
+            onClick={this.handleSortByLength}
+          >
+            Sort by length
+          </button>
+
+          <button
+            type="button"
+            className={this.state.isReversed === true
+              ? 'button is-warning'
+              : 'button is-warning is-light'}
```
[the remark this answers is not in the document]
mvjl000: BTW I think it would look much better + if some className changed, you would only correct it in one place ;)

```suggestion
            className={`button is-warning ${!this.state.isReversed ? 'is-light' : ''}`}
```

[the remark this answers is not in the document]
mvjl000: You could apply it to all other similar cases

src/App.tsx
```diff
@@ -31,65 +31,115 @@ export function getReorderedGoods(
...
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods = visibleGoods.sort((a, b) => {
+      return a.length - b.length;
+    });
+  }
+
+  if (isReversed) {
+    visibleGoods = visibleGoods.reverse();
```
mvjl000, line 49: `reverse` mutates the array, pick your fighter ;)

```suggestion
    visibleGoods.reverse();
    
    // OR
    
    visibleGoods = [...visibleGoods].reverse()
```

src/App.tsx
```diff
@@ -31,65 +31,115 @@ export function getReorderedGoods(
...
+  }
 
-  // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
```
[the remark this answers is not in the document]
WiktoriaBielec: console.log is needed for testing

PR #1886, opened 2023-12-06
solution

aureliagorzna:
- [DEMO LINK](https://aureliagorzna.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


adrian200002 (approved):
Code works well, so i think it's okay, good job.

mvjl000 (changes requested):
In this task, you were supposed to use class component. The next task is about solving it with hooks (functional component).
Refactor it to a class component.

I also suggest changing the state. Keep in state only two things - sortType and isReversed
(It was said in one of the comments not to store the goods in the state)

aureliagorzna:
I am unable to find a solution better than using the state for the goods array, the state seems to be a natural superior method to update the content, I cannot think of anything better without sacrificing the code quality or readability

mvjl000:
> I am unable to find a solution better than using the state for the goods array, the state seems to be a natural superior method to update the content, I cannot think of anything better without sacrificing the code quality or readability

@aureliagorzna Keep only `sortType` and`isReversed` in the state. When you cloned the repo, you had there a `getReorderedGoods` function that takes goods, sortType and isReversed as arguments and returns properly sorted items. Simply invoke that and render it in your app

src/App.tsx
```diff
@@ -15,81 +15,151 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
-enum SortType {
-  NONE,
-  ALPHABET,
-  LENGTH,
+type SortCallback = (a: string, b: string) => number;
+
+type SortType = 'length' | 'alphabetical' | '';
```
mvjl000, line 20: You were supposed to use enum for this

mvjl000 (changes requested):
I left a response comment above

JoyCoffeeAddict (changes requested):
Resolve previous issues, especially about `getReorderedGoods`

src/App.tsx
```diff
@@ -16,80 +16,151 @@ export const goodsFromServer = [
...
+          return -1;
+        }
+
+        if (a > b) {
+          return 1;
+        }
+
+        return 0;
+      });
+      break;
```
JoyCoffeeAddict, line 47: alphabetical is the default sort in JS, it would be enough to write `visibleGoods.sort()`

src/App.tsx
```diff
@@ -16,80 +16,151 @@ export const goodsFromServer = [
...
+      visibleGoods.sort((a: string, b: string) => {
+        if (a.length < b.length) {
+          return -1;
+        }
+
+        if (a.length > b.length) {
+          return 1;
+        }
+
+        return 0;
```
JoyCoffeeAddict, line 59: Alwyas simplify your code.

```suggestion
        return a.length - b.length
```

src/App.tsx
```diff
@@ -16,80 +16,151 @@ export const goodsFromServer = [
...
-          Reset
-        </button>
-      </div>
+interface State {
+  isReversed: boolean
+  sortType: SortType
+}
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
```
JoyCoffeeAddict, line 76: You don't have to type that 

```suggestion
  state = {
```

src/App.tsx
```diff
@@ -16,80 +16,151 @@ export const goodsFromServer = [
...
-        >
-          Reset
-        </button>
-      </div>
+interface State {
+  isReversed: boolean
+  sortType: SortType
+}
+
+export class App extends React.Component<{}, State> {
```
JoyCoffeeAddict, line 75: There is a differenece between an empty object and null in typescript. In this case we use null to say that there aren't any props. 
```suggestion
export class App extends React.Component<null, State> {
```

JoyCoffeeAddict (changes requested):

mbruhler (approved):

PR #1887, opened 2023-12-06
Solution

MarcinLigmann:
[DEMO LINK](https://MarcinLigmann.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -63,7 +63,7 @@ state = {
   sortType: SortType.NONE,
 }
 
-Sortalphabetically = () => {
+sortAlphabetically = () => {
   this.setState({ sortType: SortType.ALPHABET });
 }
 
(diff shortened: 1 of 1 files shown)


mosowiecka (approved):
Well-conceived!

src/App.tsx
```diff
@@ -26,70 +26,110 @@ type ReorderOptions = {
...
+  sortType: SortType,
+};
+
+export class App extends React.Component<{}, State> {
+state = {
+  isReversed: false,
+  sortType: SortType.NONE,
+}
+
+Sortalphabetically = () => {
```
mbruhler, line 66: functions should be written camelCase

mbruhler (changes requested):
Your solution is correct, but style needs some changes:
- functions must be written camelCase so Sortalphabetically should be sortAlphabetically

PR #1888, opened 2023-12-06
solution

adrian200002:
[DEMO LINK](https://adrian200002.github.io/react_list-of-goods/)


Timi2424 (approved):
code is fine, but it seems css is not imported into component you created

mbruhler (approved):

PR #1893, opened 2023-12-07
solution

MarekSax:
[DEMO](https://mareksax.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -36,60 +36,112 @@ export function getReorderedGoods(
 
   // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
+  if (sortType === SortType.ALPHABET) {
+    visibleGoods.sort((a, b) => a.localeCompare(b));
+  }
+
+  if (sortType === SortType.LENGTH) {
+    visibleGoods.sort((a, b) => a.length - b.length);
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
+  // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
   return visibleGoods;
 }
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -35,61 +35,113 @@ export function getReorderedGoods(
...
-        >
-          Reset
-        </button>
-      </div>
+type State = {
+  isReversed: boolean,
+  sortType: SortType,
+};
+
+export class App extends React.Component<{}, State> {
```
volodymyr-soltys97, line 63: Use a functional component for this task

src/App.tsx
```diff
@@ -35,61 +35,113 @@ export function getReorderedGoods(
...
+  render(): React.ReactNode {
+    const visibleGoods = getReorderedGoods(goodsFromServer, this.state);
+    const { sortType, isReversed } = this.state;
+
+    return (
+      <div className="section content">
+        <div className="buttons">
+          <button
+            type="button"
+            className={`button is-info ${sortType === SortType.ALPHABET ? '' : 'is-light'}`}
```
volodymyr-soltys97, line 105: Use a classnames library for add classes with condition

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/App.tsx
MarekSax: This task should be done with class component because this chapter of React course doesn't teach about hooks. Next chapter has a very similar [task](https://github.com/mate-academy/react_list-of-goods-hooks) which needs to be done using functional component and hooks.

nazarmatsevych (approved):
Great work!

PR #1895, opened 2023-12-08
Solution

Sibunas:
[DEMO LINK](https://sibunas.github.io/react_list-of-goods/)


mvjl000 (changes requested):
I think, that you were supposed to do it with a class component. The next task is about solving it with hooks (functional component).

mpiskadlo (approved):

PR #1899, opened 2024-02-21
Solution

wsadocha:
[DEMO LINK](https://wsadocha.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -26,57 +26,93 @@ type ReorderOptions = {
...
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
+        <button
+          type="button"
+          className="button is-danger is-light"
+          style={{
+            display: sortType !== 0 || isReversed ? 'inline' : 'none',
+          }}
```
BogdanMaliuta, line 100: Use conditional rendering, not display

src/App.tsx
```diff
@@ -26,57 +26,93 @@ type ReorderOptions = {
...
+      visibleGoods.sort((goodA, goodB) => goodA.localeCompare(goodB));
+      break;
+
+    case 2:
+      visibleGoods.sort((goodA, goodB) => goodA.length - goodB.length);
+      break;
+
+    default:
+      break;
+  }
```
BogdanMaliuta, line 50: Use another cases like, ALPHABET, etc., you already have Enum for this

src/App.tsx
```diff
@@ -26,57 +26,93 @@ type ReorderOptions = {
...
+    isReversed,
+  });
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info ${sortType !== 1 && 'is-light'}`}
```
BogdanMaliuta, line 73: Use library ClassNames, for conditions in classes, its more readable

BogdanMaliuta (changes requested):
Good job!
Great work but cant approve because of cond redering, and sort types

src/App.tsx
```diff
@@ -26,57 +27,103 @@ type ReorderOptions = {
...
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {sortType !== 0 || isReversed ? (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortType(0);
```
yevhenii-pyl, line 110: Will you remember in a few days what sortType 0 means? :) 

It's better to explicitly pass meaningful value like `SortType.NONE`

src/App.tsx
```diff
@@ -26,57 +27,103 @@ type ReorderOptions = {
...
 ) {
-  // To avoid the original array mutation
-  const visibleGoods = [...goods];
+  let visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  switch (sortType) {
+    case SortType.NONE:
```
yevhenii-pyl, line 37: Probably, you can combine this case with default. 

Not sure that you want to just break on a default case. I guess we may assume that default === no sorting here. 
However, it's up to you.

PR #1919, opened 2024-04-04
Implement goods sorting functionality

olafchuszno:
[DEMO LINK](https://olafchuszno.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,6 +15,11 @@ export const goodsFromServer = [
   'Garlic',
 ];
 
+type State = {
+  isReversed: boolean;
+  sortType: SortType;
+};
+
 enum SortType {
   NONE,
   ALPHABET,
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,69 +15,141 @@ export const goodsFromServer = [
...
   LENGTH,
 }
 
-type ReorderOptions = {
-  sortType: SortType;
-  isReversed: boolean;
-};
-
-// Use this function in the render method to prepare goods
 export function getReorderedGoods(
```
yevhenii-pyl, line 29: Let's use switch statement here, as sorting options may grow - so will your if-else structure do. 
switch is just a more generic way to approach it.

src/App.tsx
```diff
@@ -15,69 +15,141 @@ export const goodsFromServer = [
...
+    this.setState((state: State) => {
+      return {
+        ...state,
+        isReversed: false,
+        sortType: SortType.NONE,
+      };
+    });
+  };
+
+  sortGoodsAlphabet = () => {
```
yevhenii-pyl, line 64: Instead of having sorting methods for alphabet && length, can't you create just one, that will receive the sorting type as an argument?

src/App.tsx
```diff
@@ -15,69 +15,141 @@ export const goodsFromServer = [
...
+      return {
+        ...state,
+
+        isReversed: !state.isReversed,
+      };
+    });
+  };
+
+  render() {
+    const { sortType, isReversed } = this.state;
```
yevhenii-pyl, line 93: 👍

src/App.tsx
```diff
@@ -15,69 +15,141 @@ export const goodsFromServer = [
...
+
+    const reorderedGoods = getReorderedGoods(goodsFromServer, {
+      sortType: sortType,
+      isReversed: isReversed,
+    });
+
+    function goodsSortedOrReversed(
```
yevhenii-pyl, line 100: Better to call it 'displayResetButton' - it will just provide more clarity for functionality behind this function. 
Btw, just to go a bit fancy, can't it be an arrow function?

yevhenii-pyl (changes requested):
Nice!
Just a few adjustments and we are ready to go. 

Just out of curiosity, why have you decided to go with the class based component instead of a functional?

BogdanMaliuta (approved):
Well done!

PR #1922, opened 2024-04-22
Solution

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -32,9 +32,21 @@ export function getReorderedGoods(
   { sortType, isReversed }: ReorderOptions,
 ) {
   // To avoid the original array mutation
-  const visibleGoods = [...goods];
+  let visibleGoods = [...goods];
 
   // Sort and reverse goods if needed
+  if (sortType === SortType.NONE) {
+    visibleGoods = visibleGoods;
+  } else if (sortType === SortType.ALPHABET) {
+    visibleGoods = visibleGoods.sort((a, b) => a.localeCompare(b));
+  } else if (sortType === SortType.LENGTH) {
+    visibleGoods = visibleGoods.sort((a, b) => a.length - b.length);
+  }
+
+  if (isReversed) {
+    visibleGoods = visibleGoods.reverse();
+  }
+
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -32,9 +32,21 @@ export function getReorderedGoods(
   { sortType, isReversed }: ReorderOptions,
 ) {
   // To avoid the original array mutation
-  const visibleGoods = [...goods];
+  let visibleGoods = [...goods];
 
   // Sort and reverse goods if needed
```
yevhenii-pyl, line 37: Let's use `switch` statement here. if-else will quickly become enormous as your codebase grows.

src/App.tsx
```diff
@@ -47,37 +59,118 @@ export function getReorderedGoods(
...
+};
+
+export class App extends React.Component<{}, State> {
+  state: State = {
+    isReversed: false,
+    sortType: SortType.NONE,
+  };
+
+  reverseOrder = () => {
+    // console.log('click');
```
yevhenii-pyl, line 79: clean up :)

src/App.tsx
```diff
@@ -47,37 +59,118 @@ export function getReorderedGoods(
...
+  sortReset = () => {
+    this.setState(() => {
+      return {
+        isReversed: false,
+        sortType: SortType.NONE,
+      };
+    });
+  };
+
+  isStateDifferent = () => {
```
yevhenii-pyl, line 113: Probably, we 'stateChanged' will be more explicit naming here.

yevhenii-pyl (changes requested):
Good job!
Let's make a few code style adjustments and we are good to go.

src/App.tsx
```diff
@@ -32,9 +32,25 @@ export function getReorderedGoods(
...
   // Sort and reverse goods if needed
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+    default:
+      // eslint-disable-next-line no-console
```
yevhenii-pyl, line 46: In this case, it will be better just to return visibleGoods. 
No need to disable eslint rule.

yevhenii-pyl (approved):
Nice :) 
But try not to disable eslint rules, almost always there is a better(cleaner) approach.

PR #1923, opened 2024-05-09
list of goods solution

ZupaFly:
[DEMO LINK](https://zupafly.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -34,9 +34,20 @@ export function getReorderedGoods(
   // To avoid the original array mutation
   const visibleGoods = [...goods];
 
-  // Sort and reverse goods if needed
-  // eslint-disable-next-line no-console
-  console.log(sortType, isReversed);
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+    default:
+      break;
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
 
   return visibleGoods;
 }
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -48,34 +59,62 @@ export function getReorderedGoods(
...
+  const [sortType, setSortType] = React.useState(SortType.NONE);
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info 
+            ${sortType !== SortType.ALPHABET ? 'is-light' : ''}`}
```
anastasiia-tilikina, line 71: pls use classnames library for dynamic classes

src/App.tsx
```diff
@@ -48,34 +59,62 @@ export function getReorderedGoods(
...
-          Reset
-        </button>
+        {(isReversed || sortType !== SortType.NONE) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortType(SortType.NONE);
+              setReversed(false);
+            }}
```
anastasiia-tilikina, line 102: let's move this to a separate function

anastasiia-tilikina (changes requested):
Well done!

Just a couple suggestions here)

src/App.tsx
ZupaFly: Done

src/App.tsx
ZupaFly: i am not sure that i get you right. Some googling shows me that first i need to use `npm install classes` and then my code should be like that:
`<button
   type="button"
   className = {classNames('button', 'is-info', {'is-light': sortType !== SortType.LENGTH})}
          onClick={() => setSortType(SortType.LENGTH)}
 >
    Sort by length
 </button>` 
is it correct?

anastasiia-tilikina: exactly!) 👍🏼

anastasiia-tilikina (approved):
Great! Well done 🔥

PR #1925, opened 2024-05-22
Task solved

meljaszuk:
[DEMO LINK](https://meljaszuk.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,6 +38,21 @@ export function getReorderedGoods(
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
 
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort((a, b) => a.localeCompare(b));
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => a.length - b.length);
+      break;
+    default:
+      break;
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
+
   return visibleGoods;
 }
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -38,6 +38,21 @@ export function getReorderedGoods(
   // eslint-disable-next-line no-console
   console.log(sortType, isReversed);
```
yevhenii-pyl, line 39: Should be cleaned up

src/App.tsx
```diff
@@ -48,34 +63,59 @@ export function getReorderedGoods(
...
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(isReversed || sortType !== SortType.NONE) && (
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
```
yevhenii-pyl, line 100: A cleaner approach would be to create a separate handler for this.

src/App.tsx
```diff
@@ -48,34 +63,59 @@ export function getReorderedGoods(
...
 
       <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(goodsFromServer, { sortType, isReversed }).map(
```
yevhenii-pyl, line 112: It works just fine, but, as a rule of thumb, it's more preferable to create a variable like reorderedGoods as the result for invoking getReorderedGoods function and use it here.

yevhenii-pyl (approved):
Clean and readable!
Just a few things to adjust for future, but nothing critical. Good job! 👍

PR #1931, opened 2024-09-02
Develop

keex1k:
[DEMO LINK](https://keex1k.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -32,7 +32,25 @@ export function getReorderedGoods(
   { sortType, isReversed }: ReorderOptions,
 ) {
   // To avoid the original array mutation
+
   const visibleGoods = [...goods];
+
+  switch (sortType) {
+    case SortType.ALPHABET:
+      visibleGoods.sort();
+      break;
+    case SortType.LENGTH:
+      visibleGoods.sort((a, b) => {
+        return a.length - b.length;
+      });
+      break;
+    default:
+      break;
+  }
+
+  if (isReversed) {
+    visibleGoods.reverse();
+  }
 
   // Sort and reverse goods if needed
   // eslint-disable-next-line no-console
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -48,34 +66,59 @@ export function getReorderedGoods(
...
+  const [isReversed, setReversed] = React.useState(false);
+  const [sortType, setSortType] = React.useState(SortType.NONE);
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info ${sortType !== SortType.ALPHABET ? 'is-light' : ''}`}
```
Zibi95, line 77: There is library in every react project that allows you to write conditional classNames. Take a look into that lib documentation for future reference. https://www.npmjs.com/package/classnames

`"classnames": "^2.5.1"`

src/App.tsx
```diff
@@ -48,34 +66,59 @@ export function getReorderedGoods(
...
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {getReorderedGoods(goodsFromServer, { sortType, isReversed }).map(
+            (good, index) => (
+              <li key={index} data-cy="Good">
```
Zibi95, line 117: never use array index as a key prop
https://legacy.reactjs.org/docs/lists-and-keys.html#keys

src/App.tsx
```diff
@@ -48,34 +66,59 @@ export function getReorderedGoods(
...
+          className={`button is-warning ${!isReversed ? 'is-light' : ''}`}
+          onClick={() => setReversed(p => !p)}
+        >
           Reverse
         </button>
 
-        <button type="button" className="button is-danger is-light">
-          Reset
-        </button>
+        {(isReversed || sortType !== SortType.NONE) && (
```
Zibi95, line 99: create variable for that so the jsx part stays clean.

src/App.tsx
```diff
@@ -48,34 +66,59 @@ export function getReorderedGoods(
...
+          <button
+            type="button"
+            className="button is-danger is-light"
+            onClick={() => {
+              setSortType(SortType.NONE);
+              setReversed(false);
+            }}
```
Zibi95, line 106: To keep the JSX clean, it's a good idea to move certain logic or event handlers out of the JSX and into their own functions or handlers.

Zibi95 (changes requested):
Hi, I've added some comments to the code. Please address them, and we'll be all set.

src/App.tsx
```diff
@@ -48,34 +67,71 @@ export function getReorderedGoods(
...
+
+  const vResetButton = isReversed || sortType !== SortType.NONE;
+
   return (
     <div className="section content">
       <div className="buttons">
-        <button type="button" className="button is-info is-light">
+        <button
+          type="button"
+          className={`button is-info ${alphabetLight}`}
```
Zibi95, line 90: You can use it like that 
```suggestion
className={classNames('button is-info', {'is-light': sortType !== SortType.ALPHABET})}
```

src/App.tsx
Zibi95: fix key use `good` variables

Zibi95 (changes requested):
Hey. There is still some things that you didn't fixed.

PR #1932, opened 2024-09-19
Develop

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -70,12 +70,11 @@ export const App: React.FC = () => {
 
       <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {goodsFromServer.map((good, index) => (
+            <li key={index} data-cy="Good">
+              {good}
+            </li>
+          ))}
         </ul>
       </ul>
     </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+// import cn from 'classnames';
 
 export const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -35,47 +36,110 @@ export function getReorderedGoods(
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [isReversed, setIsReversed] = useState(false);
+
+  const reorderedGoods = getReorderedGoods(goodsFromServer, {
+    sortType,
+    isReversed,
+  });
+
+  //Just playing around with classnames
+  const basicClassesForButton = cn('button');
```
zibi-dev, line 71: If you don't apply any conditional classes then you don't have to use classnames lib

src/App.tsx
```diff
@@ -35,47 +36,110 @@ export function getReorderedGoods(
...
+    'is-light': sortType !== SortType.LENGTH,
+  });
+
+  const buttonReverse = cn(basicClassesForButton, 'is-warning', {
+    'is-light': !isReversed,
+  });
+
+  const buttonReset = cn(basicClassesForButton, 'is-danger', {
+    'is-light': sortType === SortType.NONE && !isReversed,
+  });
```
zibi-dev, line 87: You can do it in JSX code.

src/App.tsx
```diff
@@ -35,47 +36,110 @@ export function getReorderedGoods(
...
       <ul>
         <ul>
-          <li data-cy="Good">Dumplings</li>
-          <li data-cy="Good">Carrot</li>
-          <li data-cy="Good">Eggs</li>
-          <li data-cy="Good">Ice cream</li>
-          <li data-cy="Good">Apple</li>
-          <li data-cy="Good">...</li>
+          {reorderedGoods.map((good, index) => (
+            <li key={index} data-cy="Good">
```
zibi-dev, line 139: Again, the key is being used as an index 😢. I believe I mentioned this in a yours other task. Please use unique values instead. It's not a problem in this case, though.

Here is article why key is important https://www.epicreact.dev/why-react-needs-a-key-prop

Zibi95 (approved):
LGTM! Read comments left in code for reference.

